PR #1, opened 2019-11-14
tnfsBoot: make it work

mozzwald:
Revert to original sio_get_command without skip, change debug prints so they look better in serial monitor, change offset order in seek to make it work. success!


tschak909:
Fantastic!

PR #101, opened 2020-02-02
Extending bytecodes to swap.

damosan314:


damosan314:
Added support for documented 3 byte instructions.  A remapper probably doesn't have to worry about zeropage instructions (?).

PR #182, opened 2020-04-04
strip leading path items

jeffpiep:
I ran out of steam so this is not tested. Idea is to look for last '/' and send the string that follows.


jeffpiep:
fixed by Thom

PR #183, opened 2020-04-04
create custom board spec

TheMontezuma:
I was looking what "esp32doit-devkit-v1" actually means.
There is a json configuration file with the hardware specification of a specific dev board.
For example:
https://github.com/platformio/platform-espressif32/blob/master/boards/esp32doit-devkit-v1.json
It includes hardware configuration, like "f_cpu": "240000000L" (since the proper CPU frequency is defined here, it does not need to be overwritten in paltformio.ini, as I initially thought).

I actually downloaded the file with the hardware specification used so far in Fujinet project.
I renamed the file to fujinet_board.json and included it in the project.
This way it could be easily adapted to the project specific hardware.


tschak909:
AWESOME. GREAT IDEA! THANKS FOR DOING THIS! :)
-Thom

PR #186, opened 2020-04-07
Atari 820 embedded fonts (normal and sideways)

jeffpiep:
upload SPIFFS again to get the two new font files in the file system.


jeffpiep:
two other things: changed printer to 820; turned off verbose debug in TNFS.

PR #194, opened 2020-04-15
change flash partitions to bring more space for 4MB hardware

TheMontezuma:
Introduced 2 partition configurations:
- 4MB (without OTA)
- 16MB (with OTA)
Both configurations support 576kB spiffs.

Changes in paltformio.ini :
- it uses now fujinet_partitions_4MB.csv
- Bluetooth is enabled


TheMontezuma:
See:
https://docs.platformio.org/en/latest/platforms/espressif32.html#partition-tables
https://github.com/espressif/arduino-esp32/tree/master/tools/partitions

PR #206, opened 2020-04-22
moved more WIFI code from modem.cpp to fnWiFi.cpp

kaelef:


kaelef:
forgot and included my PLATFORMIO.INI in this commit - I assume you can skip it when merging?

tschak909:
it's okay. don't worry about it. :)

PR #213, opened 2020-04-28
Printer update - now with PNG screen dump

jeffpiep:
moved reset printer call out of emulator class into sio printer class
added PNG printer to the mix (html, enum, etc.)


jeffpiep:
also moved printer prototype x86 code to printer-proto directory

PR #231, opened 2020-05-13
Return error if try to download in middle of print job

jeffpiep:
The webserver returns an error if the user tries to download the print job if the printer is busy.


jeffpiep:
also did an auto format run

PR #278, opened 2020-06-20
update 4MB partition table for more space

jeffpiep:


jeffpiep:
and fix Atari 825 font for missing '~'

PR #358, opened 2020-09-12
MIDIMaze tweaks; disable B button if reboot detected early

kaelef:


kaelef:
Try again after merging last upstream

Issue #371, opened 2020-10-09
Problems connecting with BBSs


mleair:
Hi,

I received my Fujinet from The Brewing Academy yesterday. Great device overall, however, I am having trouble with BBSs.

When I connect to Atari BBSs, I find that there are either intermittent delays in receiving characters or lock ups.

My Fujinet is at version 0.5.77c13818 2020-10-05 01:27:16

I tried a couple different term programs that used the Altirra R: Handler that works with the Fujinet. I first tried the modem-programs.atr from atari-apps.irata.online. I tried both Bobterm and ICE-T.  I also tried ICE-T and Bobterm from tnfs.atari8bit.net. The same thing happens with all these term programs.

The BBSs I tested include:

basementbbs.ddns.net:9000
ptbbs.ddns.net:8000
southernamis.ddns.net:23
alcatrazbbs.ddns.net:9000
bfbbs.no-ip.com:8888

It did not matter if I logged on using ATASCII or ASCII. I would still see intermittent delays and/or lock ups. Once the device locks up, I have to reboot to use it again. Also Fujinet's webserver does not come up in a web browser when the device locks up.

It does not matter which baud rate I use on the term program. I still experience these problems.

I first reported this to Marlin at The Brewing Academy. He does not believe it is a hardware problem and suggested I file a bug.

After some experimentation, I found that I had a little bit of success if I adjust Fujinet's HSIO setting to 9 and set the term program's baud rate to 2400. However, I still occasionally get delays and lock ups when I do this.  Alcatraz BBS will always lock up after you login. I think it's due to the complex ATASCII welcome screen.

I reproduced this problem on two different computers, so I suspect it's not a problem with the computer. I reproduced it on an Ultimate 1MB Atari 800XL and a 64K 600XL. A friend of mine also reproduced the problem on his Fujinet with a stock Atari 800XL. So, it does not appear to be dependent on the computer nor the individual Fujinet since there's at least two users who have seen the same problem.

Let me know if I need to provide additional information. Hopefully this is a simple problem to fix.

Best,

Mark

mleair:
Also, I should add that this functionality appeared to work in January of this year when this youtube video was made. Note that there's no problems logging onto the Basement BBS in this video using Fujinet.

https://youtu.be/k8-XBbteakQ

-Mark

tschak909:
Logged and looking into it.

mleair:
@tschak909 

[...]

PR #411, opened 2020-12-14
Disable CONFIG_LWIP_TCP_QUEUE_OOSEQ

mleair:
For issue #371 

Disable CONFIG_LWIP_TCP_QUEUE_OOSEQ (which is used to queue incoming out-of-order segments) since it causes Fujinet to lockup when connected to Altirra based BBSs.


tschak909 (approved):
I'm Thom Cherryhomes, and I approve this message..erm..patch.

PR #488, opened 2022-03-20
More Apple 2 work and reboot state logic changed for Adam

jeffpiep:


jeffpiep:
well ... i meant to PR the CONFIG ... but this should be fine, too.

PR #494, opened 2022-05-06
Return the basename of the mounted image rather than leftmost characters of the full path.

frachel:
Most mounted images are in subdirectories on the TNFS server/SD card which causes their display on the Atari side config program to be truncated since it can only be 35 characters.

Asked some users for opinions and they agreed the actual image name is the most important thing. Barring the ability to send longer names and/or more info about the file (a future enhancement), this now returns the 'basename()' of the mounted image instead of the leftmost 35 characters.


frachel:
Closing - going to finish up my other enhancement of sending more data to the Atari about the mounted image so we can do the "basename" and/or display full path info on the config side of things.

tschak909:
Okay cool, keep us posted :)

PR #513, opened 2022-09-24
fixes #512

### LINKED ISSUE #512: [fnjson] will not return values larger than 2147483647.
Currently values larger than 2147483647, return 2147483647. If number is larger, return the double.

DaBaDa79:
(*) [fnjson] if numeric value with fraction returns as 10 digit double, otherwise returns 64 bit integer
(*) refactoring - [fnjson] uses getter method instead of directly accessing fields 
(+) optimizing [fnjson] print debugging output when getting JSON values

### DIFF
--- a/lib/fnjson/fnjson.cpp
+++ b/lib/fnjson/fnjson.cpp
@@ -9,6 +9,8 @@
 
 #include <string.h>
 #include <sstream>
+#include <math.h>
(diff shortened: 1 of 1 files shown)


tschak909 (approved):

tschak909:
Thank you! :)

Issue #514, opened 2022-09-25
Readable UTF-8 characters in JSON results


DaBaDa79:
When parsing JSON documents containing UTF-8 characters (like diacritic letters as é or the German umlauts Ä, Ö, Ü, ä, ö, ü, ß) these are emitted via SIO unmodified. End user programs could profit if the FujiNet converts all these chars to the corresponding - as available - equivalents of the destination platform (i.e. the "International Character Set" on ATARI)

PR #515, opened 2022-09-25
Json stuff

DaBaDa79:
#514 
Now the JSON Query routine can convert UTF-8 special characters to ATARI compatible ones. You (the user) can control the conversion via SIO AUX2 when issuing the "Q" command. The lowest two bits decide what to do.
Bit 0=0 - no conversion at all
Bit 0=1 - conversion dependent on Bit 1
Bit 1=0 - conversion of the common latin diacritics (á, à, é etc.) by stripping the accent (á and à to a etc.) and German umlauts (ä, ö, ü, ß etc.) to the two letter transcription (ä to ae, ß to ss etc.)
Bit 1=1 - conversion of all characters contained in the ATASCII international character set (see [Wikipedia](https://en.wikipedia.org/wiki/ATASCII#International_Character_Set))

Example in BASIC: Call until now:
`XIO ASC("Q"),#1,12,0,"N:/path/to/string/element"` and an UTF-8 **ä** will be garbage

Change it to:
`XIO ASC("Q"),#1,12,1,"N:/path/to/string/element"` and the same **ä** will become an **ae**

Change it to:
`XIO ASC("Q"),#1,12,3,"N:/path/to/string/element"` and the same **ä** will become an ATASCII **ä** (don't forget to `POKE 756,204` to change the font)

_Note 1:_ An AUX2 value of 2 has no effect as in this case bit 0 is 0 so bit 1 is ignored.
_Note 2:_ The ATASCII font has no German **ß** so bit 2 makes no difference for the conversion, **ß** will always become **ss**
_Note 3:_ As in the past AUX2 was normally zero there should be no compatibility issues.
_Note 4:_ Due to my limited C++ knowledge the code regarding the multiple nested string arrays with different size should be refactored. It is working, but ugly and redundant.

### DIFF
--- a/lib/device/sio/network.cpp
+++ b/lib/device/sio/network.cpp
@@ -1091,7 +1091,7 @@ void sioNetwork::sio_set_json_query()
 
     inp = strrchr((const char *)in, ':');
     inp++;
-    json.setReadQuery(string(inp));
+    json.setReadQuery(string(inp), cmdFrame.aux2);
     json_bytes_remaining = json.readValueLen();
     tmp = (uint8_t *)malloc(json.readValueLen());
     json.readValue(tmp,json_bytes_remaining);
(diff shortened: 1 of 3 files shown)


tschak909:
THANK YOU, SO MUCH!
-Thom

PR #517, opened 2022-10-03
[runcpm] Fix Direct I/O read bug.

tschak909:
Upgrade RunCPM to 5.8.

_kbhit() was always returning zero due to min(). (This function only has substantial use when Direct I/O is used, so it has only now been found), The fix was to revert it.


tschak909:
merging RunCPM 5.8 upgrade into master.

Issue #448, opened 2021-05-02
Problem accessing APETIME when booting a program over SIO2SD with high speed IO


billkendrick:
When booting the APOD viewer client (2021-05-01 version), either on an ATR disk image (containing MyDOS and the program) or as an XEX directly, using an SIO2SD with high speed I/O (the config on mine shows `sio high speed: 59659.08 ($08)`), the APOD client cannot _ever_ properly fetch the current time & date via FujiNet's APETIME protocol.

Unlike what happens when either (a) FujiNet is detached or turned off, or (b) the APOD client is booted so quickly that FujiNet is not yet up and running (I generally use my The Ultimate Cart SD cartridge, as it's the quickest way to fire up the XEX), which seems to be APETIME protocol coming back with an empty string, it instead seems to return some kind of gibberish. (What I've seen the APOD viewer client parsing it as were either "2000-00-15", or "2000-120-15".)

Based on this, I added extra sanity checks in APOD client (https://github.com/FujiNetWIFI/fujinet-apps/commit/f0d7212d32d2e5dda4466dd266cde77abb967f48).

PR #523, opened 2022-11-05
libssh implementation

markjfisher:
Here's my merge of the https://github.com/tnn2/esp-idf-libssh code to implement libssh.

It's mostly a straight copy of src and include dirs.

One issue I had was that libssh uses the include of "config.h", but because of the way the project is setup that includes the libsmb2 generated one, as the include path has:
```
/path/to/fujinet-platformio/components/libsmb2/include/esp
```
which means anyone doing
```
#include "config.h"
```
gets the file under smb2. sad-panda.

So I manually changed all those to "libssh/config.h"

I've merged the "compat" files that are in the esp-idf-libssh directly into the tree, rather than having it separate. This could be changed, but again is complicated by there being multiple "config.h" files all over the place.
Instead, I merged compat/config.h into libssh/config.h

There are several files not moved across, because they were not referenced in https://github.com/tnn2/esp-idf-libssh/blob/master/components/libssh/CMakeLists.txt so that build wasn't using them anyway.
Investigating, most are openssl requiring files, but it would be worth getting a second pair of eyes on this.

Note, as in the original SSH.cpp, there is no host validation nor config reading, which leaves the implementation a little fragile to MITM attacks etc.

I did try to enable pthread, but ssh init failed and I couldn't debug it, I was getting weird issues of the unicode sequence for FFFD being spammed to the monitor console whenever I add any debug to init.c, so I stopped trying to get pthread support after an hour of problems.

There are very few changes to the original c files in libssh. I've included a diff file between them below.
This is only for the files I copied into the project.
[libssh-diffs.txt](https://github.com/FujiNetWIFI/fujinet-platformio/files/9944412/libssh-diffs.txt)
This was generated with the script:
```
#!/bin/bash

SRC=/home/markf/dev/personal/atari/fujinet/esp-idf-libssh/components/libssh/libssh/src/
TGT=/home/markf/dev/personal/atari/fujinet/fujinet-platformio/lib/libssh/src

find $SRC -name \*.c\* | while read c
do
  JUST_FILE=${c#$SRC}
  TGT_FILE=${TGT}/${JUST_FILE}
  echo "-------------------------------- comparing $JUST_FILE"
  diff -u ${SRC}${JUST_FILE} $TGT_FILE
done
```

I haven't delete the libssh2 dir, but I have commented out the use of esp32sshclient in modem.h


markjfisher:
This is to fix #448

tschak909:
Thank you so much!

Issue #433, opened 2021-02-09
Refactor MODEM code


tschak909:
The current WiFi MODEM code lives monolithically inside sio/modem.cpp.

This should be re-factored outward into its own library under lib/ so it can be used by other peripheral buses (and by CP/M)

The idea here is to:

* Seperate out the core modem code
* Call an initialization function with a pointer to the desired UART.
* Replace explicit calls to fnUARTSIO to use this pointer.
* Make lib/sio/modem.cpp use the now factored library.

brad-colbert:
It this simply a "move and create an interface" job?

tschak909:
yeah.

PR #528, opened 2022-11-20
move modem into lib/ and refactor to use it with new uart member

markjfisher:
This is an attempt to match the list of things to do on #433 

Can I get some feedback on if it is what was required or not please?


tschak909:
Yeah, that makes sense. Thanks!

PR #529, opened 2022-11-21
Split fnConfig.cpp into separate files

markjfisher:
Preliminary work to split the huge file into chunks so it's easier to work on the parts.


tschak909:
Thank you! :)

PR #545, opened 2023-01-07
Add APETIME configuration to UI [#542]

### LINKED ISSUE #542: Problem accessing APETIME when run along with using APE
Hi,

when FujiNet and APE are used together and APETIME SIO call is performed both devices answer.

I suggest following options:

- config option for APETIME function in FujiNet (power-cycle-proof) that allows for [permanent off | permanent on | auto-soft-disable]
- Like FujiNet keeping quiet if it detects a remote D1: answering upon boot it could soft-disable APETIME in this case

markjfisher:
This should close #542 

New UI configuration under "Time Zone" added to allow user to disable in built APETIME functionality.

image


XeAtari:
Thank you very much for adding this configuration option.
I installed the latest firmware only this morning.
I switched APETIME to "No" and klicked "Save", but FujiNet still replied to the command.
A coldstart did not change the behaviour.

tschak909:
@XeAtari i've changed the behavior to not add the device to the bus at all, if disabled. 
The commit is here: 
https://github.com/FujiNetWIFI/fujinet-platformio/commit/e9805ac11acae78965cdbd00618fba07ab4d928c

XeAtari:
Hello,

I have installed firmware 0.5.13741956 and I can confirm that APETIME 
ON/OFF works.

(Changing from ON to OFF or from OFF to ON requires a powercycle or FN 
reset.)

Thanks a lot for implementing this function.

Kind regards,

XeAtari

==

PR #548, opened 2023-01-21
Implement eject command, disk switched status reporting and write protected status reporting.

wschaub:
This is an experimental implementation of support for the eject command, 
disk switched status and treating read only mounted disks as proper write protected disks.  

You should be able to swap out disks during the System 6.0.1 install process with this installed.


tschak909:
Thank you! :)

PR #551, opened 2023-02-01
Eject improvements

wschaub:
Change iwm so that inactive devices can still report back status/reply to init requests, this is important for removable media because
eject is removing the disk from the drive, not yanking the drive from the bus. 

this also fixes a problem where I was leaking memory when mounting a new disk on top of an already active slot
because I wasn't fully calling _disk->unmount() before destroying the _disk object and continuing with the mount process.


tschak909:
Thank you!

PR #566, opened 2023-03-17
This change makes the smartport respect the number of bytes requested

TechCowboy:
Previous to this patch, the fujinet would always return 512 bytes of data, regardless as to what was requested.


tschak909:
Thank you!

PR #578, opened 2023-03-26
Reliability fixes

wschaub:
This addresses two issues that kept me from doing a full test run of installing gsos from scratch, verifying it and adding extra applications on top while doing many disk swaps from the web UI on a single boot of the fujinet.  

the two problems encountered were: 

- webUI: memory leak that lead to sdmmc errors after enough file navigation was done.
- interrupts allowed in both CPU cores leading to watchdog timeouts obtaining locks held in the opposite core.


tschak909:
Thank you!

PR #581, opened 2023-04-05
Add support for latest espressif32 platform

oliverschmidt:
Some/many/most/all current FujiNet target systems require the espressif32 platform version 3.4.0 (with ESP-IDF version 4.3.1). 

However, at least some new target systems will require a (way) more recent ESP-IDF. One of the reasons is that newer ESP32 variants like the ESP32-S3 are only fully supported by the (most) recent ESP-IDF. As of today there are already two projects looking into running FujiNet on the ESP32-S3.

So far, the idea was to lift at some point all of FujiNet to a recent ESP-IDF. But given the number of existing target systems and the pure amount of changes Espressif has introduced with the ESP-IDF 5.x (https://docs.espressif.com/projects/esp-idf/en/v5.0.1/esp32/migration-guides/release-5.x/index.html) this seems quite a challenge :-(

Therefore this pull request introduces a new approach: FujiNet can be built both with the espressif32 platform 3.4.0 *__AND__* the espressif32 platform version 6.1.0 (with ESP-IDF version 5.0.1).

The goals of this pull request:
* Don't introduce any regression when building with platform 3.4.0
* Successfully build at least for one target system (FujiApple) with platform 6.1.0

It was tried hard to bring all FujiNet functionality to the platform 6.1.0 - there was nothing "just commented out". Anyhow, FujiApple built with platform 6.1.0 does *__NOT__* work correctly at this time.

Maybe the contributors for the existing target systems will make their code work with platform 6.1.0, maybe they don't. Maybe it's even no possible to do at all.

But as mentioned above, the major goal of this pull request is *__NOT__* to bring platform 6.1.0 to the existing target systems, rather its goal is to bring platform 6.1.0 to target systems not brought up yet.

Note: In order to make this hybrid approach work successfully it seems obligatory to have one platform 6.1.0 build be part of the FujiNet PlatformIO CI - otherwise it will break about every day!


tschak909:
Looks good to me.

oliverschmidt:
@mozzwald, @jeffpiep: This PR already has a merge conflict by now. Don't get me wrong but I don't see me resolving merge conflicts "forever". I already did so continuously in the past weeks. If you want it, merge it.

jeffpiep:
I appreciate the work you put into it. However we can’t merge into master
code that will break it. So, we need to have a plan to stop development and
make the merge work. Is there another way?

mozzwald:
> @mozzwald, @jeffpiep: This PR already has a merge conflict by now. Don't get me wrong but I don't see me resolving merge conflicts "forever". I already did so continuously in the past weeks. If you want it, merge it.

Sorry I have not had time to look at this PR. It comes at a time when we are all preparing for VCFE. My main concern is whether or not this breaks building with 3.4.0. I will make time this weekend to look deeper.

jeffpiep:
@oliverschmidt - I see your merge note "Anyhow, FujiApple built with platform 6.1.0 does NOT work correctly at this time." I take that to mean the if we merge, the FujiApple firmware on master will no longer work and will have to be fixed.

oliverschmidt:
> @oliverschmidt - I see your merge note "Anyhow, FujiApple built with platform 6.1.0 does NOT work correctly at this time." I take that to mean the if we merge, the FujiApple firmware on master will no longer work and will have to be fixed.

That's a major misunderstanding. FujiApple still builds and works fine with platform 3.4.0. It's just not *__additionally__* working with platform 6.1.0.

jeffpiep:
OK, thank-you for clearing that up!

oliverschmidt:
My concern is that the migration to IDF V5 might turn out (way) more complex than anticipated. Therefore this PR offers a way to not have to stop everything, migrate everything and then restart everything. I put actually quite some effort into making this possible. It would have been *__much__* simpler for me to just create a PR that makes FN *__only__* work with IDF V5.

oliverschmidt:
> [...] My main concern is whether or not this breaks building with 3.4.0. I will make time this weekend to look deeper.

That's of course *__THE__* question! I'd appreciate (a lot) if you could find the time to assess that - and consider a merge in case the result is positive.

So far I was only able to get feedback from an FujiApple owner saying that he couldn't see any issue with the 3.4.0 build :-)

jeffpiep (approved):
now that i understand the changes, looks good. thanks for the PR!

mozzwald:
I was able to build and boot Atari FujiNet with Platform 3.4.0. Tried building 6.0.1 for Atari and it failed. Applied several fixes but got stopped when linking. Anyway, more work needs done on the newer platform but it the framework looks good. :+1:

oliverschmidt:
Thanks - I really hope that there are in fact no hidden regressions!

PR #582, opened 2023-04-07
Allow removal of external tristate buffer on FujiApple

jeffpiep:


jeffpiep:
To use, need to bypass the tristate buffer by connecting MOSI to RDDATA. Only works with SPI Fix - separate SP and SD SPI busses. Set a global NO3STATE define in platformio.ini to build.

PR #607, opened 2023-05-23
[iwm] ignore bad data packets on control code 0

wschaub:
Sometimes on GS/OS shutdown or switching betwen ProDOS and GS/OS a control code 0 data packet gets sent with garbage and a bad checksum.

This patch accepts bad data packets when the last recieved command is a control command with code 0.


lebraseric:
Thank you :-)

PR #609, opened 2023-06-02
rc2014: build: update to platformio esp platform 6.1

juzzas:
The update to platformio ESP32 platform causes a couple of new compiling issues. Fix them.


tschak909:
Thank you! :)

PR #623, opened 2023-06-28
Fix BUILD_COCO ifdef and other BUILD_ comments

markjfisher:
lib/device/drivewire had an incorrect ifdef to ATARI still for cassette code, causing it to be picked over sio.

other BUILD_ comments also fixed.


tschak909:
Thank you!

PR #624, opened 2023-06-30
Add return value for tcpServer.begin() and have all callers use return

markjfisher:
Clean up error handling of tcpServer().begin


markjfisher:
fixing...

markjfisher:
I seem to have lost some file changes when merging master into my line. fixing

PR #626, opened 2023-07-04
Fix sending all bytes of file and don't abort if ATN is asserted immediately after EOI

FozzTexx:
There are 3 fixes:

* Send all bytes of a file, including the last one
* ATN being asserted immediately after the EOI byte isn't an error, still return the final byte
* ATN being asserted right after acking the EOI byte should not release DATA since interrupt already asserted it


tschak909:
Thank you!

PR #630, opened 2023-07-09
rc2014: modem: refactor to use command frames

juzzas:
Since we moved to using an SPI bus for RC2014, the modem support had
become unusable. Refactor the the modem to use command frames,
adding a circular buffer to emulate a UART. Code on the RC2014 will
batch characters together to send/receive command frames in the
"driver"


tschak909:
Thank you! :)

PR #631, opened 2023-07-10
Remove guard introduced for saving passwords

markjfisher:
This should fix the issue with not saving wifi passwords.


tschak909:
Thank you!

PR #632, opened 2023-07-17
printer updates for A2/A3 for SP buffering

robjustice:
Hopefully it won't break the Apple2 side of things. This is for handling buffering of the printer output better.


tschak909:
Thank you!

PR #634, opened 2023-07-24
Fix calculation of the buffer size to store a filename in when callin…

frachel:
…g the ellipsize utility function.


tschak909 (approved):
Looks good, to me.

PR #635, opened 2023-07-24
_fnDisks is declared with MAX_DISK_DEVICES, which changes sizes.

JagMod:
This loop was assuming MAX_DISK_DEVICES = 8, which is not the always the case.


tschak909:
Thank you! :)

PR #643, opened 2023-07-26
cppcheck warnings fixed

JagMod:
A couple of these are 'todo' comments, the author should look at it to determine the best fix.


tschak909:
Thank you!

tschak909:
I will look at these when I get the cycles, to see what best to do.

PR #648, opened 2023-07-27
cppcheck findings: passing std::string by const ref whenever possible

JagMod:


tschak909 (approved):
looks good. Thank you.

PR #651, opened 2023-07-31
Add -z option to build.sh to build firmware zip

markjfisher:
For @mozzwald 

run:
```shell
$ ./build.sh -z
```

it will clean, build filesystem, then build with the zip options enabled.
cleans up after itself.

It works by finding the line containing "; post:build_firmwarezip.py" and then inserts the post option ABOVE it.
After building zip, it removes the option it added, leaving the platformio.ini file as it was.

It will fail if there's no line with the comment at the end of the extra_scripts section, so the current template should work.


mozzwald:
Awesome, Thanks!

PR #655, opened 2023-08-15
RC2014: fuji: align SSID size with other platforms

juzzas:
The RC2014 does not match the other platforms with the size of the SSID array. Fix this.


tschak909:
Thank you!

PR #659, opened 2023-08-20
RC2014: fuji: implement base64 commands

juzzas:
Implement the base64 encode/decode commands.

This has only been compile-tested.


tschak909:
Thank you! :)

PR #660, opened 2023-08-20
RC2014: network: remove need for N: specifier

juzzas:
Remove the need for the N: specifier for opening
the network device.

This is part of the rc2014 library refactoring for the network device.


tschak909:
Thank you!

PR #661, opened 2023-08-20
Mac - just merge to not fall behind

jeffpiep:


jeffpiep (commented):
i reviewed to make sure core file changes were benign. All are compiler directives  in "hardware" so I can use the FujiApple Rev00 SPI fix hardware.

PR #662, opened 2023-08-21
RC2014: fuji: implement hashing commands

juzzas:
Implement the SHA1/SHA256/SHA512 commands.

This has only been compile-tested.


tschak909:
Thank you!

PR #663, opened 2023-08-22
RC2014: network: fix json query

juzzas:
Since we removed the N: prefix from network device specs for RC2014, we missed one usage in rc2014_set_json_query()!

Fix this.


tschak909:
Thank you!

PR #665, opened 2023-08-26
IEC - BASIC command format standardization, get app keys working, json line endings when returning object

EricCarrGH:


tschak909:
Thank you! :)

PR #670, opened 2023-09-06
Add extended Adapter Config information

markjfisher:
This extends command $E8 (get adapter config) to return additional strings for the various IP and MAC values to save the client from having to do it.


tschak909:
Thank you!

PR #671, opened 2023-09-13
Add bad-connection detection to set-ssid

markjfisher:
This change will first try the given ssid/password combination before saving it to fnconfig.

It works by shutting down the existing wifimanager, starting a new mini version with some blocking code that checks whether a disconnect or connect event happens.
The disconnect tells us the config was bad, and we then stop the saving, and report the error to the host.

If the config was good, the wifi manager is restarted, with a small delay to allow it to connect, and the host is then given a success code, allowing them to then check the wifi status and see everything is good.

I've also (sorry) add the start of the block directory code to this commit as I will be adding it anyway. Nothing calls it, but I kept losing it in between stashing changes.

I've test compiling everything only on Atari board.

### DIFF
--- a/lib/config/fnc_wifi.cpp
+++ b/lib/config/fnc_wifi.cpp
@@ -57,8 +57,7 @@ void fnConfig::_read_section_wifi(std::stringstream &ss)
     Debug_println("Reading wifi section");
 
     // Throw out any existing data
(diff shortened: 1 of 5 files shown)


tschak909:
Thank you!

PR #677, opened 2023-10-09
Add 00 dstats $FB command to change json parameters

markjfisher:
See https://github.com/FujiNetWIFI/fujinet-platformio/wiki/N:-SIO-Command-$FB---Set-JSON-Parameters

This allows host to change JSON parameters from the host using ioctl commands:

```cpp
#ifdef BUILD_ATARI
    err = network_ioctl(0xFB, 0, 1, ipify_json, 0, 0, 0);    // set json Query Param
    err = network_ioctl(0xFB, 1, 0x00, ipify_json, 0, 0, 0);    // set json Line Ending (default is usually 0x9b)
#endif
```


tschak909:
Thank you!

PR #679, opened 2023-10-22
Add .DO support. Look inside .DSK files to detect sector order.

mike-wiese:
For issue #616
- add support for DOS order .DO disk images
- look for ProDOS volume header info inside .DSK image files to detect if they are in DOS order or ProDOS order

### DIFF
--- a/lib/device/iwm/disk.cpp
+++ b/lib/device/iwm/disk.cpp
@@ -515,17 +515,33 @@ mediatype_t iwmDisk::mount(FILE *f, const char *filename, uint32_t disksize, med
   }
 
   // Determine MediaType based on filename extension
-  if (disk_type == MEDIATYPE_UNKNOWN && filename != nullptr)
+  if (disk_type == MEDIATYPE_UNKNOWN && filename != nullptr) {
     disk_type = MediaType::discover_mediatype(filename);
+    if (disk_type == MEDIATYPE_DSK) {
+        // determine DO or PO based on file contents
+        disk_type = MediaType::discover_dsk_mediatype(f, disksize);
+    }
+  }
 
   if (deviceSlot < 4) // SP drive
   {    
     switch (disk_type)
     {
+    case MEDIATYPE_DO:
+        Debug_printf("\r\nMedia Type DO");
+        _disk = new MediaTypeDO();
+        _disk->_media_host = host;
+        _disk->_mediatype = disk_type;
+        strcpy(_disk->_disk_filename, filename);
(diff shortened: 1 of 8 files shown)


tschak909 (approved):
Thank you!

lib/device/iwm/disk.cpp
```diff
@@ -535,8 +551,6 @@ mediatype_t iwmDisk::mount(FILE *f, const char *filename, uint32_t disksize, med
           readonly = false;
```
oliverschmidt, line 551: Why is the section above not duplicated to MEDIATYPE_DO?

mike-wiese: `high_score_enabled` is only ever set to true for .2MG files in MediaTypePO::mount

oliverschmidt: I see. Thanks for the quick and helpful answer :-)

Another question I couldn't find an answer for easily: Does the new code make sure that Disk II mounting of PO media is only an option for 140kB files?

mike-wiese: I admit I don't really understand the high score code. I don't recall anything like that for other floppy emulators, so I assume it is a FujiNet thing?

The new code is not checking for 140kB files. I was thinking to allow up to 160kB, or anywhere between 35-40 whole tracks, since WOZ format supports up to 40. Not common but pretty easy. I could pass back `MEDIATYPE_UNKNOWN` from the ::mount calls if the size isn't supported.

tschak909:
On the Apple2, The High score code simply checks for a write request (on
something currently mounted as read-only), and if the underlying file
permissions allow it, it will honor it by re-opening the file as
read-write, writing the block, and then reopening as read-only. This is so
that high score tables can be written to disk.

-Thom

lib/device/iwm/disk.cpp
oliverschmidt: I'm not sure how to interpret your answer. Do you say that there's already code in place that checks for some size or do you say that you plan to add such code?

mike-wiese: The new code calls through to the old DSK code … as far as I can tell the DSK code does not check the size.

I am proposing to add file size checks, and support up to 40 tracks while I'm in the code.

tschak909:
Go fro it.

-Thom

lib/device/iwm/disk.cpp
oliverschmidt: Thanks for the explanation :-)

PR #680, opened 2023-10-24
fix web mount list on apple2

mike-wiese:
Fix web interface mount list to not show (D�:) on every drive slot.

The return type of `get_disk_id` is `int`, but httpServiceParser (the only caller) casts it to `char`, and then checks if it is > 0. Unfortunately `char` is unsigned, so that didn’t work as intended.

Fixed by changing `get_disk_id` to return 0 instead of -1.


mike-wiese:
This same fix might be needed in mac/iwmFuji, but I'm not set up to test that so I left it alone.

tschak909:
Thank you!

PR #681, opened 2023-10-28
Remove esp32s3_platform_version.

oliverschmidt:
Now that esp32_platform_version is 6.x there's no more need for a esp32s3_platform_version.


tschak909:
Makes sense, thank you!

PR #683, opened 2023-10-28
Remove dependency on os.h

oliverschmidt:
The FN code using this code presumes anyhow that 'os_malloc' is the same as 'malloc'.


tschak909:
Thank you!

PR #684, opened 2023-11-04
apple2: Support 40 track 5.25 inch disk images

mike-wiese:
- support 36 and 40 track .DO, .PO, and .DSK files
- check file size is supported size
- plumb MEDIATYPE_UNKNOWN errors in mount code


tschak909:
Thank you!

PR #708, opened 2024-03-17
Updated certificate file

michaelsternberg:
Fujinet-pc build for ATARI was unable to read files from github.com. Instead the mgHTTP client get call was returning a TLS error for invalid certificate.


tschak909:
Thank you!

PR #709, opened 2024-03-21
Adjusted sio_set_prefix() to comply with NCD documentation

michaelsternberg:
Tested with both NOS and N-enabled DOS XL. Assumption is other versions use same N-tools wand will behave the same.


tschak909:
Thank you!

PR #710, opened 2024-03-28
add long-press A triggers mount_all for FujiMac

benjamink:
Long-pressing the `A` button will trigger a `mount_all` for FujiMac devices.


tschak909:
Thank you!

PR #711, opened 2024-03-31
[coco] Set Drivewire BAUD rate based on ROM image selection

djtersteegc:
Read A14/A15 ROM dipswitch via IO36/IO39


tschak909:
Thank you! :)

PR #712, opened 2024-04-04
Support for TCP protocol in TNFS client.

trekawek:
To connect to a tnfsd server supporting TCP, add `_tcp.` prefix to its hostname, eg.:

```
_tcp.apps.irata.online
```


tschak909:
Thank you! :)

PR #713, opened 2024-04-06
Connect to TCP by default and fallback to UDP.

trekawek:
With this PR, the firmware will first try to connect to tnfsd using TCP protocol. If the connection is not possible (tcp/16384 doesn't respond) or the connection is accepted but server doesn't respond to the MOUNT command (because it's an old server version that doesn't handle TCP commands), the client will fallback to UDP.

My suggestion is to wait with merging this PR until these two servers support TCP:

* fujinet.atari8bit.net
* fujinet.pl

I already contacted owners.


tschak909:
Thank you!

PR #714, opened 2024-04-07
Switch from TNFS_PROTOCOL_UNKNOWN to TCP or UDP after a transaction.

trekawek:
Before this commit, we were switching from TNFS_PROTOCOL_UNKNOWN to TNFS_PROTOCOL_TCP after successfully sending a message. If the reply didn't come back, we switched back to TNFS_PROTOCOL_UDP.

After this commit, we keep TNFS_PROTOCOL_UNKNOWN until there's a successful reply (when we switch to TNFS_PROTOCOL_TCP) or a failure (then we go to TNFS_PROTOCOL_UDP).

This is mostly a cleanup. There's one minor change in behaviour: after this commit, if the user provides _tcp. prefix and the TCP connection fails, client *won't* fallback to UDP, as it would before the commit.


tschak909:
Thank you! :)

PR #717, opened 2024-04-20
Fixed a clang error by initializing a variable separately from its declaration.

boisy:
The error was: variable-sized object may not be initialized


tschak909 (approved):

tschak909:
Thank you! :)

PR #719, opened 2024-04-22
Hitting RESET on the CoCo causes an OP_RESET transaction code to occur

boisy:
This resets the system to the configuration disk image.


tschak909:
Thank you!

PR #727, opened 2024-05-05
Enable session recovery in TNFS.

trekawek:
This PR enables the session recovery in the TNFS lib. Before that, this feature was only active outside of the ESP_PLATFORM.


tschak909:
Thank you!

PR #744, opened 2024-05-26
Use recursive mutex for the transactions.

trekawek:
The standard mutex is not reentrant and we use recursive calls to _tnfs_transaction in the session recovery (which will block with std::mutex).


tschak909:
Thank you!

PR #746, opened 2024-05-31
add step to fix name for upload

benjamink:
Fix firmware file output name to something standard so the `deploy-nightly` workflow step can find it when build targets don't match matrix item names.  

Also removed the date stamp from the file name for Andy.

NOTE: need to manually delete the old asset files with the date from the [Nightly Builds](https://github.com/FujiNetWIFI/fujinet-firmware/releases/tag/nightly) Release since the new files will have different names so won't overwrite current ones.


tschak909:
Thank you!

PR #749, opened 2024-06-06
Try to recv next message if the receive seq is smaller than expected.

trekawek:
This way we'll read all the messages from the buffer until we get the right one.

Before this commit we were sending another request, which would keep the send/recv pair out of sync.


tschak909:
Thank you!

PR #751, opened 2024-06-19
Adam GET_TIME now returns the same format as apple GET_TIME

TechCowboy:
Previous format:
day,month,year,hour,min,sec
New format:
century, year,month,day,hour,min,sec


tschak909:
Thank you!

PR #753, opened 2024-07-13
Added directory filtering when filter pattern ends with `/`

hcgernhardtiii:
The CoCo repository is replete with directories for each application/game/whatever, and as such being able to filter the directory names as well is helpful.  This commit adds this functionality by adding a trailing `/` to the filter string.


tschak909:
Thank you. :)

PR #754, opened 2024-07-13
Updated SDFAT, FTP, and SMB filesystems to recognize and utilize directory filtration.

hcgernhardtiii:
I don't have devices that use the other filesystems, unfortunately…


tschak909:
Thank you!

PR #755, opened 2024-07-15
[fnWifi] re-enable mdns for Atari

nwah:
Re-enable mDNS in Atari build so that the web interface can be accessed at e.g. `http://fujinet.local` without any further configuration.


tschak909:
Thank you!

PR #756, opened 2024-07-15
[build] use json.dumps to properly escape firmware release.json

nwah:
This avoids an edge case where commit messages might contain characters that create invalid JSON and prevent flashing.
Screenshot 2024-07-14 at 8 30 18 PM

**BEFORE**
Screenshot 2024-07-14 at 8 27 45 PM

**AFTER**
Screenshot 2024-07-14 at 8 24 29 PM


tschak909:
Thank you!

PR #758, opened 2024-07-25
[WIP] [fnDNS] lowercase hostname before resolving ipv4

nwah:
I don't know how 2 c++

But might not hurt to do a lowercase here right before resolving the IP, to avoid the gotcha with people putting stuff in all caps


nwah:
Closing for now. Seems like bug is limited to `.local` hostnames, so can probably fix in more targeted way for lwIP's minimal mDNS implementation

PR #759, opened 2024-07-30
support for ESP32-S3 and migration to new ADC and timer interface

gtortone:
Hi,

this PR contains initial support for ESP32-S3 and new ADC and timer ESP-IDF interface.


tschak909:
Thank you! :)

PR #761, opened 2024-07-31
unused vars and undefined structure fields

gtortone:
Hi,
due to some misleading warnings during build I propose to remove unused variables and disable warning related to uninitialized structure fields for some source files using #pragma compiler feature;


tschak909:
Thank you!

PR #769, opened 2024-08-22
Change SP_RDDATA to be the GPIO pin that is connected to the Disk II.

FozzTexx:
This renames a couple of pin constants so they better reflect what they represent. SP_RDDATA is now the pin that will be controlled by SPI & RMT and SP_RD_BUFFER is the pin that is used to enable/disable the buffer chip on older FujiNet Apple II revisions.


tschak909:
Thank you!

PR #772, opened 2024-08-24
Fixes a few minor warnings

FozzTexx:
Fixes a few minor warnings to make it easier to find where the real errors are.

* Declare fallthrough is expected
* Add missing fields to struct init in iwm_ll.cpp
* Resolve volatile warnings


tschak909 (approved):
Yup, I'm okay with this.

PR #773, opened 2024-08-26
Adjusted malloc for PC build for ATARI

michaelsternberg:
Modified following files to _**not**_ use PSRAM functions to allocate memory when building for the fujinet-pc for ATARI.
- `lib/device/sio/fuji.cpp`
- `lib/device/sio/fuji.h`


tschak909:
Thank you!

PR #775, opened 2024-08-27
Enable using Lolin D32 Pro as Apple II FujiNet.

FozzTexx:


mozzwald:
Couple of issues with this...

1. creation of the separate `common.h` for D32 boards is unnecessary. To reduce confusion, just create a new D32 board for apple 2.
2. The commit was tested on 3 variants: Rev1 (works), Rev000 with SPIFIX and No Tristate buffer (works), Rev0 without SPIFIX (broken). Below is debug output from the failed boot. My guess is that it's related to `lib/hardware/fnsystem.cpp` hardware checks, but I have not looked that deeply at the PR changes. If I figure it out, I will let you know. Until then, we can leave this open.

```FujiNet v1.4 2024-08-07 20:49:13 (APPLE) Started @ 7
Starting heap: 4357656
[src/main.cpp:135] main_setup(): Heap: 212351

FujiApple SPI fix not found
Detected Hardware Version: Rev0
Button A Enabled on IO0
Button B Disabled
Button C (Safe Reset) Enabled on IO14
LittleFS mounted.
SD mounted.
fnConfig::load
Load fnconfig.ini from SD
fopen = /fnconfig.ini : ok
fnConfig::load read 1175 bytes from config file
Reading wifi section
FLASH Config Storage: Enabled
fnConfig::load read 1175 bytes from FLASH config file
iwmNetwork::iwmNetwork()

Config General Boot Mode: 0
FileSystemLittleFS::filehandler_open /autorun.po rb
new FileHandlerLocal
disk MOUNT /autorun.po

Media Type POFileHandlerLocal::read

IWM FujiNet based on SmartportSD v1.15

FujiNet Hardware timer started
RMT configured for Disk ][ Output
SPI configured for smartport I/O
IWM GPIO configuredAvailable heap: 4182528
Setup complete @ 167 (160ms)
[src/main.cpp:436] main_setup(): Low Heap: 168759

WiFiManager::start() complete
WiFiManager attempting to connect:
ssid = retrostead
WiFi connect attempt to SSID "notyomamawifi"
WIFI_EVENT_STA_STARTesp_wifi_connect returned 0

WIFI_EVENT_STA_CONNECTED
IP_EVENT_STA_GOT_IP
Obtained IP address: 192.168.1.41
SNTP client start
Starting web server on port 80
SNTP time sync event: Tue Aug 27, 17:17:28 2024 +0000

Reset
Reset Cleared
en35Host = 0
[lib/bus/iwm/iwm.cpp:77] print_packet(): packet:
FCFFFFFF00C381808080808281808582


handling init command
Sending INIT Response Packet...
Send REQ timeout
Send REQ timeout
Send REQ timeout
Send REQ timeout
Send REQ timeout
Drive: 81
```

FozzTexx:
> 1. creation of the separate `common.h` for D32 boards is unnecessary. To reduce confusion, just create a new D32 board for apple 2.

It made sense to me because these pins will never be changed when using a D32 Pro. I have removed the d32-common.h and reverted iec-d32pro.h.

> 2. The commit was tested on 3 variants: Rev1 (works), Rev000 with SPIFIX and No Tristate buffer (works), Rev0 without SPIFIX (broken). Below is debug output from the failed boot. My guess is that it's related to `lib/hardware/fnsystem.cpp` hardware checks, but I have not looked that deeply at the PR changes. 

That file wasn't changed, so it shouldn't have broken. Did you try to boot a couple of times? The SmartPort stuff has been flakey for me to get it to boot. I sometimes have to try several times before it gets past the initial setup sequence and starts reading blocks.

mozzwald:
> That file wasn't changed, so it shouldn't have broken. Did you try to boot a couple of times? The SmartPort stuff has been flakey for me to get it to boot. I sometimes have to try several times before it gets past the initial setup sequence and starts reading blocks.

I see it was not changed. There is hardware configuration that happens in `fnsystem.cpp` when fujinet boots and detects which pull ups are present on the actual hardware and you've changed some defines which may or may not be affecting this detection. I'm not sure if this is the issue but was just my first guess.

I did try multiple boots without success. My next step is to flash current master and test it to ensure it's still working with this hardware. The other 2 hardware variants worked fine with your PR

FozzTexx:
> I did try multiple boots without success. My next step is to flash current master and test it to ensure it's still working with this hardware. The other 2 hardware variants worked fine with your PR

It might be line 877 (old) / 881 (new) in iwm_ll.cpp. It looks like it used to be doing the equivalent of `IWM_BIT_SET(SP_RD_BUFFER)` and I changed it to `IWM_BIT_INPUT(SD_RD_BUFFER)` probably because the comment said it was switching it to hi-z.

mozzwald:
> It might be line 877 (old) / 881 (new) in iwm_ll.cpp. It looks like it used to be doing the equivalent of `IWM_BIT_SET(SP_RD_BUFFER)` and I changed it to `IWM_BIT_INPUT(SD_RD_BUFFER)` probably because the comment said it was switching it to hi-z.

Current master works so the issue is from the PR. I tried changing this as you suggested, but it did not work either.

FozzTexx:
> I tried changing this as you suggested, but it did not work either.

Try also changing line 868 (new) in iwm_ll.cpp from IWM_BIT_OUTPUT to IWM_BIT_CLEAR.

mozzwald:
> Try also changing line 868 (new) in iwm_ll.cpp from IWM_BIT_OUTPUT to IWM_BIT_CLEAR.

ok, with both changes it works on all hardware so if you update PR with that I think it's all good

FozzTexx:
> update PR with that

Done.

PR #778, opened 2024-09-10
Move IWM byte decoding into a function.

FozzTexx:
Moved IWM byte decoding routine from `iwm_read_packet_spi` and created `iwm_decode_byte` so that the same routine can be later re-used when decoding the data sent during Disk II write.


tschak909 (approved):
Looks good, thanks!

PR #779, opened 2024-09-10
Split up service() into sub services

FozzTexx:
Split up service() into sub services to make it easier to add more services (such as Disk II write).


tschak909 (approved):
Looks good, thank you!

PR #786, opened 2024-09-12
Fix Disk II mounting from web interface.

FozzTexx:


tschak909 (approved):
looks good.

tschak909:
Thank you!

PR #790, opened 2024-09-15
fnFTP: skip sending password if server didn't 331

nwah:
Adds small check to FTP lib to skip trying to send a password if server doesn't require one, i.e. server returns `230` in response to initial login call.


tschak909:
Thank you!

PR #794, opened 2024-09-17
Improve reliability of SmartPort packet receive

FozzTexx:
* Prepare linked list for SmartPort command packet and block packet in advance
* Directly access SPI registers to start receive and avoid overhead/latency of ESPIDF SPI driver

### DIFF
--- a/lib/bus/iwm/iwm_ll.cpp
+++ b/lib/bus/iwm/iwm_ll.cpp
@@ -4,8 +4,8 @@
 
 #include <string.h>
 
-#include "esp_rom_gpio.h"
-#include "soc/spi_periph.h"
+#include <esp_rom_gpio.h>
+#include <soc/spi_periph.h>
 
 #include "iwm_ll.h"
 #include "iwm.h"
@@ -22,6 +22,15 @@
 #define IWM_BITLEN_FOR_BYTES(bytecount, freq, buffer) ({ \
       (bytecount) * 8 * IWM_SAMPLES_PER_CELL(freq);      \
     })
+#define IWM_NUMBYTES_FOR_BITS(bitcount, buffer) ({      \
+  size_t bufbits = sizeof(*(buffer)) * 8;           \
+  size_t blen = (bitcount) *                        \
+    IWM_SAMPLES_PER_CELL(smartport.f_spirx);        \
+  blen = (blen + bufbits - 1) / bufbits;            \
+  blen;                                             \
+    })
+
+#define IWM_CHUNK_SIZE 128
 
 volatile uint8_t _phases = 0;
 volatile sp_cmd_state_t sp_command_mode = sp_cmd_state_t::standby;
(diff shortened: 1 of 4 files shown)


FozzTexx:
Simplest test procedure:

Setup a ProDOS disk with BASIC.SYSTEM and save a BASIC program called STARTUP with the contents:

```100 HOME
110 PRINT "FujiNet SmartPort packet test"
120 FOR I = 1 TO 1000:NEXT I
130 PRINT CHR$(4);"PR#5"
```

Replace the `5` in `PR#5` with whatever slot your SmartPort controller is in. Mount the disk on FujiNet and boot it and let it run for a while. Compare to master and see that master will get `I/O ERROR` within a minute or two vs. this branch which will keep booting and rebooting for quite a long time.

Pre-built firmware for Apple FujiNet from commit ID 58c26a72:

[fujinet-APPLE-58c26a72.zip](https://github.com/user-attachments/files/17080055/fujinet-APPLE-58c26a72.zip)

dillera:
You can pull down the test images from FT here:

TNFS://fujinet.diller.org/APPLE/TEST/PR794/hello.dsk
TNFS://fujinet.diller.org/APPLE/TEST/PR794/hello.po

dillera:
Issue report: running bounce world client on this fw- it stops working in about 3 min. Before on 1.4 release you could run bouncy client for 12 hours +.

robjustice:
I have the original spi-fast-start branch cloned from FozzTexx's github still here locally before the 'cleanup'. When I load that into my FN its works very well for the looping boot test. If I take the current Fujinet Master and add this PR 794 to make a branch locally for testing. That does not work any where near as well. It stops with the request timeouts after only a few mins. This is testing on a IIe with DB19 disk controller and Grapper Minus card with v6a firmware.

FozzTexx:
> I have the original spi-fast-start branch cloned from FozzTexx's github still here locally before the 'cleanup'. 

Can you make a bundle with that older commit and send me the bundle?

robjustice:
Here is a tar of the lib directory. Let me know if that's ok.
[spi-fast-start_lib.tar.gz](https://github.com/user-attachments/files/17087819/spi-fast-start_lib.tar.gz)

FozzTexx:
> Here is a tar of the lib directory. Let me know if that's ok. [spi-fast-start_lib.tar.gz](https://github.com/user-attachments/files/17087819/spi-fast-start_lib.tar.gz)

What's the commit ID of HEAD~1 so I can check that out and then extract the tar on top of it?

robjustice:
95140df1 (HEAD -> spi-fast-start, origin/spi-fast-start) Wait for transaction to finish after receiving packet.
2642dfc2 Get SPI receive to start faster.
6960d909 fnFTP: skip sending password if server didn't 331
27240d86 [iwm] protect against FN crash after a reset on an open network connection
927ae471 [iwm][fuji] Make fuji a real device so it doesn't rely on the DISK_0 hack

FozzTexx:
Cool, I'll see if I can reproduce the problem and see what the difference is.

PR #796, opened 2024-09-21
Link to build artifacts in PR comments

benjamink:
This change to the `autobuild.yml` GitHub Action will upload the build artifacts for each of the firmware builds when changes are pushed to a PR & comment in the PR with links to download the latest artifacts.  Subsequent pushes to the PR will rebuild the artifacts & the links in the persistent comment will be updated to point to the new artifacts.  This workflow will only create a single comment in a PR with the links & just keep updating that one comment so it doesn't become noisy.  

By default GitHub stores build artifacts for 90 days & then they are automatically deleted.

This is an example of how the comment should look: 
image


dillera:
This is a draft- do you want it merged- need to flip it to a real PR.

benjamink:
Converting this to ready for review.  I was tempted to tweak it more to add the short SHA to the artifact names but I think we should just merge this to see if it actually works first.

PR #799, opened 2024-09-28
Revert "Update fnHardwareTimer to use driver/gptimer.h"

FozzTexx:
Testing has proven 771 is the cause of SmartPort unreliability.

Tested revert with tnfs://ayce.fozztexx.com/SmartPort.Exorciser.po

UNADJUSTEDNONRAW_thumb_7902


dillera (approved):
Looks good, fixes master, runs all SP HTTP apps fine. Zero errors on the excorcisor.

PR #804, opened 2024-10-06
ATARI: Improve trailing slash in sio_set_prefix

michaelsternberg:
Add trailing slash in `lib/device/sio/network.cpp` in method `sio_set_prefix` to `prefixSpec_str`.
Later, adding trailing slash to `prefix` is no longer necessary as it already exists in `prefixSpec_str`.


tschak909:
Thank you!

PR #807, opened 2024-10-15
Developer tool to put commit ID into FN_VERSION_FULL

FozzTexx:
Add debug_version.py to put commit ID into FN_VERSION_FULL during build without altering include/version.h


dillera:
Looks good with it commented in and out on OSX...

Screenshot 2024-10-16 at 10 19 30 AM

PR #810, opened 2024-11-03
ATARI: Corrected 'NCD N1:' to clear prefix

michaelsternberg:
Fix to a bug I introduced in a previous change. 

NCD N1:  (by itself is intended to clear the prefix, but was setting the prefix  to '/')


tschak909:
Thank you!

PR #812, opened 2024-11-09
Change debug_version.py to not use most recent modified date

FozzTexx:
Change debug_version.py defaults to not use most recent modified date since it causes everything to rebuild.


tschak909:
Thank you!

PR #819, opened 2024-12-13
[WIP] QR commands for FujiNet device

nwah:
⚠️  WIP - NOT READY FOR MERGE ⚠️ 

Could use help on how to set things up nicely as shared library all platforms call, with some platform-specific implementations. Will probably still just be Atari for now as I don't have hardware/software to test other platforms.

Will squash everything into 1 commit when ready.

---
Adds new serial command(s) to the Fuji device for generating QR codes.

It follows the same pattern as base64/hash commands:
1. Send text to FN to be encoded, which is stored in a buffer
2. Tell FN to encode it. (For the QR code stuff you can specify version (size), and error correction level. The result is stored as bytes in a buffer.)
3. Ask the FN for the length in bytes of the result
4. Read bytes from the FN

Three output modes available: bytes (1 `0x00` or `0x01` per QR "module"), bits (1 bit per module), or ATASCII (each char represents 2x2 chunk of the QR code).

Screenshot 2024-12-13 at 2 21 31 PM


tschak909:
Thank you!

PR #820, opened 2024-12-14
URL shortener

nwah:
Adds a simple URL shortener to the webserver, primarily for use with the new QR code encoding feature.

You are pretty limited in the amount of data you can encode in a QR code if it can only be 21x21, and this may be too short for many URLs. This URL shortener allows you to take an arbitrarily long URL and shorten to `http://{fujinet_ip}/url/{id}`.

For Atari, you can tell the QR encoder to shorten the supplied URL first by setting bit 4 on AUX2 (i.e. adding 16 to the desired error correction level).


tschak909:
Thank you!

PR #825, opened 2025-01-06
Only set the network unit if the command count has a value sent by fujinet-lib that supports it

markjfisher:
When fujinet-lib was updated to support network unit for apple2, it increased the count value of bytes the command sends.

This change detects that increase to determine on an sp_read/write/status/control whether the network unit id is present.


tschak909 (approved):
Looks good. :)

PR #830, opened 2025-01-27
Updated SP-over-SLIP protocol.

oliverschmidt:
Note: This is an incompatible protocol change. It requires a companion change in AppleWin.

FujiNet makes use of command list bytes there weren't used so far. This only works with CBus because CBus alwaways tranports all possible command list bytes no matter what command is given. Now the SP-over-SLIP protocol shows exactly the same behavior CBus protocol.


tschak909:
Ok, looked over this. Looks good to me. @markjfisher can yell at me if it breaks. I let it through. 

Thank you!

PR #831, opened 2025-01-29
Validate the path in Fujinet "hosts list" browser in the web UI.

trekawek:
Before this commit it's possible to escape the configured root by specifying `..` between the URL-encoded slash: `%2F`:

```
http://localhost:8000/browse/host/1/%2F..%2F..%2F..%2F
```


tschak909:
Awesome thank you!

PR #836, opened 2025-02-10
fix uint32 compile errors

benjamink:
Fix compile time errors due to mis-matched formats in debug output functions.

image


tschak909:
Thank you!

PR #839, opened 2025-02-14
[fuji] add set_slot_hostname to rest of platforms

nwah:
add set_slot_hostname to rest of platforms and remove Atari #ifdefs for web ui host management endpoints


tschak909:
Thank you!

PR #841, opened 2025-02-18
[fuji] set_slot_hostname: repopulate config

nwah:
when updating slot hostname (from webui), also call `_populate_config_from_slots` so the updated name gets copied over there as well


tschak909:
Thank you!

PR #842, opened 2025-02-19
QR: new bitmap output + fix

nwah:
- fix bug where output formats other than bytes wouldn't work after the 1st encode in atari implementation
- new "bitmap" output option for QR encoder, that outputs things in MSB->LSB order suitable to copy directly into e.g. sprite data
- renamed `QR_OUTPUT_MODE_BITS` to `QR_OUTPUT_MODE_BINARY` and `::to_bits` to `::to_binary` to make less confusing(?) with new bitmap mode

bitmapped

### DIFF
--- a/lib/device/sio/fuji.cpp
+++ b/lib/device/sio/fuji.cpp
@@ -2136,10 +2136,6 @@ sioDisk *sioFuji::bootdisk()
 }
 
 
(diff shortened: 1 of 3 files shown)


tschak909:
Thank you,!

PR #843, opened 2025-02-19
[qr] validate params; error if input > capacity

nwah:
The underlying QR generation library silently generates invalid QR codes if the input text exceeds the capacity for the given size/error correction level.

Unfortunately there's no straightforward formula for calculating capacity, so this uses a table to validate and returns an error if not. Also adds checking for version and error correction level parameters. For Atari implementation, an SIO error is generated in this case.


tschak909:
Thank you!

PR #844, opened 2025-02-20
Support recursive filtering in TNFS.

trekawek:
There are 2 features here:

1. Filtering pattern starting with `!` will be transformed into `**/pattern*`, required by the new tnfsd TNFS_DIROPT_TRAVERSE option (see https://github.com/FujiNetWIFI/tnfsd/pull/16)
2. Add long-path support for `util_ellipsize()` (see below).

Before this PR, the `util_ellipsize()` removes the middle part of the string, replacing it with `...`, to fit into the specified `dstsize`. After this PR, the function checks if the specified string is a long path (and contains slash in the middle). If that's the case, it'll try to return the full basename of the file + as much as the parent path as possible, e.g.:

```
games/s/spy vs spy 2/spy vs spy 2.atr
```

will be translated into:

```
...spy 2/spy vs spy 2.atr
```

If the basename doesn't fit into the specified size, it'll be ellipsized on its own and prefixed by the `.../`, e.g.:

```
.../this is some very very ... long long basename.atr
```

### DIFF
--- a/lib/FileSystem/fnFsTNFS.cpp
+++ b/lib/FileSystem/fnFsTNFS.cpp
@@ -270,19 +270,25 @@ bool FileSystemTNFS::dir_open(const char * path, const char *pattern, uint16_t d
 
     if (!!pattern) {
         thepat = realpat;
-        strlcpy (realpat, pattern, sizeof (realpat));
-        if (realpat[strlen(realpat)-1] == '/') {
-            Debug_print (
-                "FileSystemTNFS::dir_open applying pattern to directories\n"
-            );
-            realpat[strlen(realpat)-1] = '\0';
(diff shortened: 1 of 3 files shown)


tschak909:
Yup, am pulling to test.

trekawek:
@tschak909, I think it's ready for review/merging now. The new code should be only called if the the filter starts with `!`  or if the filename contains `/`, so generally if the recursive search feature is used.

tschak909 (commented):
this looks ok to me. let me test.

tschak909 (approved):
Yup, beat on it for a few minutes, I'm ok with it. Let's get it in. Thank you.

PR #846, opened 2025-02-21
[qr] fix: always clear in buf even if encode fails

nwah:
Fixes edge case where after QR encoding fails, subsequent attempts also fail


tschak909:
Thank you!

PR #851, opened 2025-02-28
Feat/apple2 fujinet pc tweaks

nwah:
- make web ui drive slots for apple 2 on fujinet-pc look same as on real hardware
- add some `data-` attributes to drive slots to make it easier to hook onto them for browser extension stuff


tschak909:
Thank you!

PR #852, opened 2025-03-01
Set default baud rate for RS232 to 115200

benjamink:
Change the default baud rate for RS232 to 115200 rather than 9600


tschak909:
Thank you!

PR #860, opened 2025-03-16
RS232 ESP32-S3 Support

mozzwald:
This adds support for building ESP32-S3 RS232 firmware. Debug output is sent over the S3 USB CDC port instead of the usual USB to Serial chip on other FujiNets.

Tested with RS232-REV0 prototype board and the YD-ESP32-S3 devkit board fly wired to it.


tschak909:
cool.

FozzTexx (approved):

PR #863, opened 2025-03-18
Fix/862 don't crash on config reboot

FozzTexx:
See issue https://github.com/FujiNetWIFI/fujinet-firmware/issues/862 for details on how to test and make sure that v1.5-rc1 exhibits the problem for you and that this fixes the described problem.


dillera (approved):
Looks good to me.

PR #873, opened 2025-03-21
V1.5 memory optimizations

tschak909:
This is based on v1.5-rc and adds a configuration option which includes memory optimizations which vastly limit the amount of DMA buffers that the WiFi code captures, doubling the amount of available memory.

All features need to be tested. Especially WIFI Association with access points.


tschak909:
I think this is okay. I haven't detected any behavioral changes in atari, apple, or coco.

FozzTexx (approved):

Issue #874, opened 2025-03-22
[apple2] WDT timeout when writing from one Disk ][ image to another.


tschak909:
Setup:
Apple ][ plus with SoftSP v6 in Slot 5
Apple ][ DB19 controller in Slot 6.

Procedure:

(1) mount DOS 3.3 MASTER in S6,D1
(2) mount blank DOS 3.3 disk prepared with CiderPress in S6,D2 (I stored on my local TNFS)
(3) BOOT
(4) BRUN FID
(5) COPY FILES
(6) COPY SOURCE S6,D1
(7) COPY DEST S6,D2
(8) FILENAME =
(9) PROMPTING? NO.
(10) INSERT DISKS
(11) Wait for copy to complete

For me it failed during copy of BIORHYTHM, with the following crash:

```
19:08:41.668 > Copy track: 73
19:08:41.680 > Copy track: 69
19:08:41.691 > Copy track: 65
19:08:41.702 > Copy track: 61
19:08:41.714 > Copy track: 57
19:08:41.725 > Copy track: 53
19:08:41.736 > Copy track: 49
19:08:41.748 > Copy track: 45
19:08:41.759 > Copy track: 41
19:08:41.770 > Copy track: 37
19:08:41.782 > Copy track: 33
19:08:41.793 > Copy track: 29
19:08:41.804 > Copy track: 25
19:08:41.816 > Copy track: 21
19:08:41.830 > Copy track: 17
19:08:41.847 > Copy track: 13
19:08:41.862 > Copy track: 12
19:08:41.862 > trk pos 012 on d2
19:08:41.862 > Disk II iwm queue receive 2944 31938 34754 50304
19:08:41.863 > Disk II write Qtrack/sector: 96/10  bit_len: 2816
19:08:41.868 > Disk II used: 2944
19:08:41.868 > Disk II sector data: 10
19:08:41.868 > DSK writing track 24 sector 10 
19:08:41.997 > Copy track: 15
19:08:42.014 > Copy track: 19
19:08:42.027 > Copy track: 23
19:08:42.039 > Copy track: 27
19:08:42.051 > Copy track: 31
19:08:42.062 > Copy track: 35
19:08:42.074 > Copy track: 39
19:08:42.077 > Copy track: 40
19:08:42.077 > trk pos 040 on d2
19:08:42.077 > Disk II iwm queue receive 2944 38338 41154 50304
[...]

Issue #875, opened 2025-03-22
Quickly switching between Disk II drive 1 and drive 2 gets wrong drive data


FozzTexx:
Reported ppuskari by on discord

Steps to repeat:

1. Connect FujiNet to computer that is capable of using two Disk II drives on external port (not IIc)
2. Mount ProDOS 2.4.2 disk in drive 1 and a different ProDOS disk in drive 2
3. Boot ProDOS 2.4.2 from FujiNet
4. When Bitsy Bye loads, press the number key for the slot the Disk II controller is in (usually 6)
5. If reading the alternate drive happens before the Apple II turns off the drive motor the FujiNet will not actually switch disks and Bitsy Bye will read the catalog from the wrong disk

FujiNet appears to be ignoring drive select line because drive wasn't stopped first.

PR #877, opened 2025-03-22
[apple2] remove printfs that are causing timing issues.

tschak909:
See issue #874 . We are putting out too much debugging info, and it is marginally throwing timing. Commenting these two fixed the issue with my softSP v6 system.


lib/device/iwm/disk2.cpp
```diff
@@ -149,7 +149,8 @@ void IRAM_ATTR iwmDisk2::change_track(int indicator)
         ((MediaTypeWOZ *)_disk)->track_len(track_pos),
         ((MediaTypeWOZ *)_disk)->num_bits(track_pos),
         NS_PER_BIT_TIME * ((MediaTypeWOZ *)_disk)->optimal_bit_timing);
-    Debug_printf("\nCopy track: %d", track_pos);
+    // This printf nudges timing too much.
+    // Debug_printf("\nCopy track: %d", track_pos);
```
FozzTexx, line 153: This one was what clued me in to what's going in with https://github.com/FujiNetWIFI/fujinet-firmware/issues/875

FozzTexx (dismissed):

FozzTexx (approved):

PR #878, opened 2025-03-22
Apple2 new disk dos33

tschak909:
Much needed quality of life enhancement that adds 'D'OS 3.3 to 'N'ew disk.

To test:

* Go to any writable host slot
* Create new disk
* Select 'D'OS 3.3 for disk type 
* Enter a filename, e.g. FOO.DO
* Mount into S6D1
* Should boot to a blank ] 
* Ensure file is 143,360 bytes.
* Linux 'file' can indicate whether it is a DOS 3.3 image
* Booting and resulting catalog should show completely blank disk image VOL 254.


FozzTexx:
Creating a disk works but when I reset the FN and go back into CONFIG the new disk isn't on the selected drive anymore. It reverts to what I had on there before. This is probably a bug for a different ticket though.

FozzTexx (approved):

PR #879, opened 2025-03-22
Few small fixes

a8jan:
[coco] fuji device: check nullptr on disk image open
- check for nullptr on disk image open
- set errorCode on file (disk image) open
- set errorCode on dir open

[coco] network device open - deviceSpec from null terminated string

url parser: properly set base_name if there is no extension

[HTTP FS] fix dir_open
- use proper buffer size (COPY_BLK_SIZE) when reading data from HTTP client
- fix file size parsing
- make sure the directory url ends with '/' to avoid unnecessary redirect

fnDirCache - use pointers for sorted entries
- no need to copy directory entries when sorting/filtering

[PC] WebUI - one restart button is enough

### DIFF
--- a/data/webui/template/www/index.tmpl.html
+++ b/data/webui/template/www/index.tmpl.html
@@ -166,9 +166,7 @@
 				<header class="module-header">
 					HARD<span class="logowob"></span>WARE
 					<span style="float: right">
-						{% if tweaks.fujinet_pc %}
-						<input type="button" id="restartButton" value="Restart..." onclick="restartButton()" style="width: 7em">
-						{% else %}
+						{% if not tweaks.fujinet_pc %}
 						<form action="/config" method="post">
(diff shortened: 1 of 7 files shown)


FozzTexx:
Should this be going into v1.5-rc instead of master?

tschak909 (dismissed):

tschak909:
probably, but can't this be pulled into v1.5?

FozzTexx:
> probably, but can't this be pulled into v1.5?

It's less work to go straight into v1.5.

FozzTexx:
This should also go in as a Rebase so the individual commits aren't blended together, not sure why it's not showing that option

a8jan:
oh, sorry if I picked the wrong branch for PR

tschak909 (approved):

PR #881, opened 2025-03-22
FNFW-875: [diskii] Switch track when drive is changed without turning motor off.

### LINKED ISSUE #875: Quickly switching between Disk II drive 1 and drive 2 gets wrong drive data

FozzTexx:
Fixes #875

### DIFF
--- a/lib/bus/iwm/iwm.cpp
+++ b/lib/bus/iwm/iwm.cpp
@@ -567,6 +567,7 @@ bool IRAM_ATTR iwmBus::serviceDiskII()
       fnSystem.delay(1); // need a better way to figure out persistence
       if (iwm_drive_enabled() == iwm_enable_state_t::on)
       {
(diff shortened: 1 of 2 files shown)


tschak909 (approved):

tschak909:
Thank you!

PR #883, opened 2025-03-23
Apple2 create new fix slot fall off

tschak909:
Disks created and assigned via Create New have been falling off the slot list, because they aren't persisted.

To test:
* Select Writable host slot
* Create New DOS 3.3 image
* Give a filename
* Boot the result
* Write to disk, to verify that you can write.
* Reset the FujiNet
* PR#6
* Verify that the disk is still visible in web UI.


tschak909:
Closing this to fold into the dos 3.3 PR.

FozzTexx:
Did you actually change autorun.po for this fix or did it get messed up from merging somehow? I didn't flash the FS so I wasn't using this autorun.po, but the new disks persisted after resetting,

tschak909:
Not for this fix, but there was a conflict, so I rebuilt.

FozzTexx:
> Not for this fix, but there was a conflict, so I rebuilt.

So autorun.po shouldn't be part of this PR?

tschak909:
it shouldn't be.

FozzTexx:
> it shouldn't be.

Ok I'll fix it.

FozzTexx (approved):

PR #885, opened 2025-03-23
Remove what appears to be dead code for disabling Disk II drive 2

FozzTexx:
I can't find anything that calls these functions. This code appears to be dead. Especially since in disk2.h `disableD2()` calls `diskii_xface.enableD2()` which seems entirely backwards.


tschak909 (approved):
looks good.

Issue #884, opened 2025-03-23
[v1.5-rc] DOS 3.3 FID reports writable Disk ][ image being write protected


tschak909:
Test procedure:

* Mount DOS 3.3 in S6D1 (R)
* Create new disk for S6D2 (W)
* Boot
* SAVE FOO,S6,D2 (works)
* DELETE FOO,S6,D2 (works)
* BRUN FID,S6,D1
* COPY FILES
* Source S6D1
* Dest S6D2
* Filename =
* PROMPTING? (N)
* COPYING FILE HELLO
DISK WRITE PROTECTED

wtf?

-Thom

tschak909:
A similar issue presents itself trying to use COPY II PLUS to copy either FILES or DISK (without formatting), D2 shows up as WRITE PROTECTED.

tschak909:
Showing when it happens from the firmware side. 

```
13:04:34.345 > disk ii enable states: 00
13:07:32.476 > disk ii enable states: 01
13:07:32.476 > disk ii enable states: 01
13:07:32.476 > start diskII d1
13:07:32.628 > trk pos 018 on d1
13:07:32.640 > trk pos 019 on d1
13:07:32.644 > trk pos 020 on d1
13:07:32.649 > trk pos 021 on d1
13:07:32.653 > trk pos 022 on d1
13:07:32.657 > trk pos 023 on d1
13:07:32.660 > trk pos 024 on d1
13:07:32.664 > trk pos 025 on d1
13:07:32.667 > trk pos 026 on d1
13:07:32.671 > trk pos 027 on d1
13:07:32.673 > trk pos 028 on d1
13:07:32.676 > trk pos 029 on d1
13:07:32.679 > trk pos 030 on d1
13:07:32.682 > trk pos 031 on d1
13:07:32.685 > trk pos 032 on d1
13:07:32.688 > trk pos 033 on d1
13:07:32.690 > trk pos 034 on d1
13:07:32.694 > trk pos 035 on d1
13:07:32.696 > trk pos 036 on d1
13:07:32.699 > trk pos 037 on d1
13:07:32.702 > trk pos 038 on d1
13:07:32.705 > trk pos 039 on d1
13:07:32.707 > trk pos 040 on d1
13:07:32.710 > trk pos 041 on d1
13:07:32.713 > trk pos 042 on d1
13:07:32.716 > trk pos 043 on d1
13:07:32.719 > trk pos 044 on d1
[...]

Issue #882, opened 2025-03-23
Divide by zero while doing disk ][ write.


tschak909:
* Mounted DOS 3.3 system master in S6D1 
* Created DOS 3.3 image, placed in S6D2
* Loaded FID
* COPY FILES
* SOURCE SLOT 6 DRIVE 1
* DEST SLOT 6 DRIVE 2
* FILENAME =
* PROMPTING? NO.

During copy of FILE BIORHYTHM, I/O Error resulted, with the following debug output:

```
21:45:02.798 > trk pos 084 on d2
21:45:02.953 > Disk II iwm queue receive 3072 6886 9702 50304
21:45:02.953 > Disk II write Qtrack/sector: 84/2  bit_len: 2816
21:45:02.961 > Disk II used: 3072 3fff5618
21:45:02.961 > DSK writing track 21 sector 14 
21:45:03.073 > Disk II iwm queue receive 3072 13286 16102 50304
21:45:03.073 > Disk II write Qtrack/sector: 84/4  bit_len: 2816
21:45:03.081 > Disk II used: 3072 3fff6e38
21:45:03.081 > DSK writing track 21 sector 13 
21:45:03.178 > Disk II iwm queue receive 3072 19430 22246 50304
21:45:03.178 > Disk II write Qtrack/sector: 84/6  bit_len: 2816
21:45:03.186 > Disk II used: 3072 3fffa278
21:45:03.186 > DSK writing track 21 sector 12 
21:45:03.289 > Disk II iwm queue receive 3072 25830 28646 50304
21:45:03.289 > Disk II write Qtrack/sector: 84/8  bit_len: 2816
21:45:03.297 > Disk II used: 3072 3fffa278
21:45:03.297 > DSK writing track 21 sector 11 
21:45:03.425 > trk pos 056 on d2
21:45:03.425 > Disk II iwm queue receive 3072 31974 34790 50304
21:45:03.428 > Disk II write Qtrack/sector: 84/10  bit_len: 2816
21:45:03.434 > Disk II used: 3072 3fffa278
21:45:03.434 > DSK writing track 21 sector 10 
21:45:03.586 > trk pos 012 on d2
21:45:03.586 > Disk II iwm queue receive 3072 38374 41190 50304
21:45:03.589 > Disk II write Qtrack/sector: 84/12  bit_len: 2816
21:45:03.594 > Disk II used: 3072 3fff7a48
21:45:03.594 > DSK writing track 21 sector 9 
[...]

PR #886, opened 2025-03-23
allow ::start() to be called before ::stop() so that it can be used to change drives

FozzTexx:
Fixes https://github.com/FujiNetWIFI/fujinet-firmware/issues/884 and https://github.com/FujiNetWIFI/fujinet-firmware/issues/882


tschak909 (approved):
works.

PR #888, opened 2025-03-24
allow ::start() to be called before ::stop() so that it can be used to change drives #886

FozzTexx:
Fixes https://github.com/FujiNetWIFI/fujinet-firmware/issues/884 and https://github.com/FujiNetWIFI/fujinet-firmware/issues/882

Pointed to v1.5-rc this time. But you better double check me!


tschak909:
Yes. I see it this time.

tschak909 (approved):

PR #889, opened 2025-03-24
Revert "Revert "allow ::start() to be called before ::stop() so that it can be used to change drives""

tschak909:
Reverts FujiNetWIFI/fujinet-firmware#887 god damn it.


tschak909 (commented):
arghh!

FozzTexx:
Master currently has no difference with 1994fae5a8f519a1d77eb25f3232a91af1bc887e, which is where it should be. If you want to get rid of all the commit clutter we'll need to do a reset back to that commit.

tschak909:
Accident. Like stepping on a dog chew twice in 30 mins.

PR #896, opened 2025-03-28
Bring things up to date with platform-espressif32 v6.10.0

idolpx:
This is intended to bring us up to date to benefit from the latest fixes and enhancements in toolchain and esp-idf.

I also eliminated the use of a static version of the "vfs" component and combined other components into the standard directory.

Most of the changes to FN specific files were includes that were missing from headers.

These changes need to be tested on physical devices and confirmed working.

I did build for all target platforms with success.  
image

### DIFF
--- a/lib/bus/cx16_i2c/cx16_i2c.h
+++ b/lib/bus/cx16_i2c/cx16_i2c.h
@@ -7,6 +7,7 @@
 #include <freertos/queue.h>
 #include <utility>
 #include <string>
+#include <driver/i2c.h>
 
 #define CX16_DEVICEID_DISK 0x31
 #define CX16_DEVICEID_DISK_LAST 0x3F
@@ -258,7 +259,7 @@ private:
     /**
      * @brief I²C slave port
      */
-    int i2c_slave_port = 0;
+    i2c_port_t i2c_slave_port = I2C_NUM_0;
 
     /**
      * @brief I²C receive buffer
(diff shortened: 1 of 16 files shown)


FozzTexx:
This needs to wait until v1.5.0 is done, lets put it into draft mode. It's also way too large of a commit, it needs to be broken up into stages, with each stage being thoroughly tested before being merged.

idolpx:
Most of the changes are just from moving the components around.  That didn't change any of those files.
I broke it up into 3 commits that I thought made sense.  Please advise.

FozzTexx:
> I broke it up into 3 commits that I thought made sense.

Ok. Most of the time a PR is viewed as a single commit since it will be added with a Squash & Merge. With multiple commits that need to remain as separate commits we will need to be careful to make sure that a PR doesn't end up all rolled up into a single commit.

idolpx:
It needs all 3 of these to compile properly on 6.10.0.

FozzTexx:
Disk II read still works the same as it does on master. Disk II write seems to be running into the RAM problem that we fixed for the v1.5 release.

Can we split this up into 3 PRs? One for each commit that's in this PR? I know the VFS stuff won't have as many entries until the final commit but I don't think it really matters for a few days while we get this in.

tschak909:
Yes, we can do that.
-Thom

FozzTexx:
@idolpx can you rebase from master and drop commit 1d7c14?

idolpx:
ok...

PR #897, opened 2025-03-29
Fix 859

### LINKED ISSUE #859: Bug in displaying drives in Apple II CONFIG
> Minor display bug in Apple2 Config:
> 1. Start with empty Disk II drives
> 2. Highlight Disk II slot and press `E` to eject
> 3. Only first letter of drive label remains (`S` or `N` instead of `S6D2` or `NA`)
> 4. Tab back to host list and drive labels go back to normal
> 
> Image
> Image
> Image 

 _Originally posted by @nwah in [#858](https://github.com/FujiNetWIFI/fujinet-firmware/issues/858#issuecomment-2726999638)_

tschak909:
fixes #859 by adjusting horizontal offsets for line clear and Empty messages depending on whether disk ][ slots are selected.


FozzTexx (approved):
It fixes exactly what it was supposed to. But if I go down to the next line and push `E` then the Empty shifts over to align with the other row. Would be nice if people would stop putting in hardcoded values and instead define constants. 😜

F7D04E30-3A4D-4862-8EDE-DD2D980FBAE3_1_102_o
1B995428-57B7-4DD2-98F7-EB41703A1053_1_102_o

PR #906, opened 2025-04-05
[PC] Refactor FujiNet-PC HTTP client (rebased)

a8jan:
- process the response parts when data arrives,
    - do not wait for the entire response (MG_EV_HTTP_MSG event)
    - replace mg_http_connect() with mg_connect()
    - use MG_EV_READ event to handle all incoming data
    - allow responses larger than mongoose buffer limit (3MB)
- change protocol to HTTP/1.1
    - support for chunked transfer encoding
- add _perform_fetch() to fetch the response in parts
- implement PROPFIND method
- handle HEAD, MKCOL COPY and MOVE methods
- fix mongoose TLS buffer overflow (back-ported)


tschak909 (approved):

tschak909:
Thank you!

PR #908, opened 2025-04-07
Adds reporting disk mounted state to iwm_stat_read_device_slots()

FozzTexx:
Added bit `DISK_ACCESS_MODE_MOUNTED` and removed unused `DISK_ACCESS_MODE_FETCH`. Fixed some things where `is_config_device` wasn't being set. Removed `#ifdef ESP_PLATFORM` from `iwmFuji::setup()` since mounting the autorun.po appears identical to what's already in `insert_boot_device()`.

Also added check to make sure spi_buffer has been allocated in `iwm_sp_ll::encode_spi_packet()` since I was able to make FN Guru error while working on making a disk read test program.


tschak909 (approved):

tschak909:
Thank you

PR #911, opened 2025-04-15
rs232-s3: fix sdcard detect, send debug output to usb, enable button A, pinmap fixes

mozzwald:
for upcoming RS232-REV1 hardware using ESP32-S3:
 * Fix inverted SD Card detect pin
 * Use printf for debug output so it's sent over USB instead of UART
 * Enable button A for RS232
 * GPIO pinmap updates/fixes


FozzTexx (approved):
rev0 still works, that's about all I can test.

tschak909:
close enough for government work. thank you :)

PR #915, opened 2025-04-22
[adam][fnSystem] fixes #914 - ensure setup_card_detect() is run.

### LINKED ISSUE #914: Eject (IO15) sensing not working
Reported by John Lundy via Facebook:

It has been a while since I’ve used ADAM FujiNet or had to build more units for stock. I recently built more FujiNet Mini units, and during testing, noticed the SD card eject/insert sense is no longer functioning after booting. The card will only work when booting fresh or after a reset. I can see I/O15 working fine with a low when a card is inserted and high when removed. After doing some troubleshooting, this functionality hasn’t worked since the release of v1.4. Dropping all the way back to v1.3 will work as expected. The serial debug shows absolutely nothing when the card is removed/inserted, as if I/O15 is no longer being monitored for any state change at all in code now.

tschak909:
With some bits in fnSystem getting shuffled around, Adam no longer had the call to setup_card_detect() to handle resetting or mounting the SD card upon insertion/removal.

### DIFF
--- a/lib/hardware/fnSystem.cpp
+++ b/lib/hardware/fnSystem.cpp
@@ -1102,6 +1102,7 @@ void SystemManager::check_hardware_ver()
     */  
     _hardware_version = 1;
     safe_reset_gpio = PIN_BUTTON_C;
+    setup_card_detect((gpio_num_t)PIN_CARD_DETECT);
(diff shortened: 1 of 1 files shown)


mozzwald (approved):
that's the fix, thanks!

PR #922, opened 2025-04-29
Bring things up to date with platform-espressif32 v6.10.0 (2nd try)

idolpx:
Changes needed to compile for v6.10.  Needs to be tested on all platforms.

This PR is without the changes to components.

### DIFF
--- a/lib/bus/cx16_i2c/cx16_i2c.h
+++ b/lib/bus/cx16_i2c/cx16_i2c.h
@@ -7,6 +7,7 @@
 #include <freertos/queue.h>
 #include <utility>
 #include <string>
+#include <driver/i2c.h>
 
 #define CX16_DEVICEID_DISK 0x31
 #define CX16_DEVICEID_DISK_LAST 0x3F
@@ -258,7 +259,7 @@ private:
     /**
      * @brief I²C slave port
      */
-    int i2c_slave_port = 0;
+    i2c_port_t i2c_slave_port = I2C_NUM_0;
 
     /**
      * @brief I²C receive buffer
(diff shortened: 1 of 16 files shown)


FozzTexx:
Still getting the insufficient RAM error with Disk II write:

`14:29:54.579 > disk ii [1] enable states: 03E (89960) SPI_SMA: Insuffucient RAM 34987 < 53248 + 4992`

FozzTexx:
> Still getting the insufficient RAM error with Disk II write:
> 
> `14:29:54.579 > disk ii [1] enable states: 03E (89960) SPI_SMA: Insuffucient RAM 34987 < 53248 + 4992`

@tschak909 any ideas why we're running out of RAM again after you made all those config changes?

tschak909:
Maybe the SDK config file changes accidentally got rolled back?

FozzTexx:
> Maybe the SDK config file changes accidentally got rolled back?

The sdkconfig files aren't part of this commit or the ones that shuffled files around in `components`.

mozzwald:
do we need to figure out the bank switching to gain access to the other 4mb of psram that is not used yet?

FozzTexx:
> do we need to figure out the bank switching to gain access to the other 4mb of psram that is not used yet?

No, because it's the DMA RAM that we're in short supply of. Everything that can use the PSRAM already is. @tschak909 made a whole bunch of changes for the v1.5.0 release to reduce how much of the DMA RAM that WiFi was eating, but it seems those changes aren't enough now?

mozzwald:
ahh, I see

tschak909:
worth a check, and it wouldn't be unprecedented, i'll take a look when I can get my head together. :)

FozzTexx (approved):
Tested with latest master and I can boot Apple II DOS 3.3, `SAVE` a BASIC program, `NEW` it out, and `LOAD` it back in. SmartPort works same as it did before too.

PR #923, opened 2025-04-30
[sio] add set_baudrate command (0xEB)

gtortone:


gtortone:
- add new command set_baudrate to change UART baudrate between external host and ESP32 (useful with Pico or other MCU)
- disabling status read when called from sio_poll_interrupt to avoid pauses during data transfer with TNFS

PR #924, opened 2025-04-30
new set_baudrate command, atari2600 fixes and network protocol changes

gtortone:
- add new command set_baudrate to change UART baudrate between external host and ESP32 (useful with Pico or other MCU)
- disabling status read when called from sio_poll_interrupt to avoid pauses during data transfer with TNFS
- add some "not used" pins to atari2600.h to allow build


tschak909 (approved):
Thank you.

PR #927, opened 2025-05-07
Allocate Disk II write buffer large enough to hold only 1 sector

FozzTexx:
Since we don't support full track formatting yet, I had another idea to try to get around the RAM starvation that the PlatformIO v6.10.0 is causing. Instead of allocating enough to capture an entire track I reduced the size to be just enough to hold 1 sector. Seems to work fine.

### DIFF
--- a/lib/bus/iwm/iwm_ll.cpp
+++ b/lib/bus/iwm/iwm_ll.cpp
@@ -31,6 +31,10 @@
       blen;                                             \
     })
 
+#define PREALLOC_D2W_BUFFER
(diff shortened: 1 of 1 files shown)


FozzTexx:
Forgot to add, since the buffer is so much smaller I am making it pre-alloc the buffer during startup instead of dynamically whenever the drive spins up. I can change it if anyone thinks that's a bad idea, it's just a `#define`.

tschak909 (approved):
I'm ok with it. :)

PR #929, opened 2025-05-20
Add printer for rs232

tschak909:
Adds Printer subdevice to RS232 as sub-device 0x40.

### DIFF
--- a/lib/device/rs232/printer.cpp
+++ b/lib/device/rs232/printer.cpp
@@ -73,7 +73,7 @@ void rs232Printer::rs232_write(uint8_t aux1, uint8_t aux2)
         linelen = 20;
         break;
     default:
-        linelen = 40;
+        linelen = 1;
     }
 
     memset(_buffer, 0, sizeof(_buffer)); // clear _buffer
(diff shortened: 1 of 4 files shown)


tschak909:
Someone wanna push this through pls?

lib/device/rs232/printer.cpp
```diff
@@ -229,6 +229,8 @@ void rs232Printer::set_printer_type(rs232Printer::printer_type printer_type)
         break;
     }
 
+    _pptr->setEOL(0x0D);
```
FozzTexx, line 232: Would '\r' be clearer here? It's definitely an ASCII character being set not a magic number.

lib/printer-emulator/pdf_printer.cpp
```diff
@@ -366,6 +365,10 @@ bool pdfPrinter::process_buffer(uint8_t n, uint8_t aux1, uint8_t aux2)
         }
         else
         {
+#ifdef BUILD_RS232
+            if (c == 0x0A)
```
FozzTexx, line 369: Would '\n' be clearer here?

tschak909: sure.

Issue #930, opened 2025-05-21
RS232 baudrate config


FozzTexx:
Do something so `#define RS232_BAUDRATE` doesn't keep getting changed on every PR. How hard is it to add this as a setting in config.ini? And make it so the web UI can change it?

FozzTexx:
Can now be changed via web UI: https://github.com/FujiNetWIFI/fujinet-firmware/pull/1016

PR #931, opened 2025-05-21
Implement 930 rs232 baud rate

tschak909:
Implements #930 by adding an [RS232] section:

```
[RS232]
baud=9600
```


FozzTexx (approved):
Tested it and it seems like it works.

PR #933, opened 2025-05-22
fixes #932 - DO NOT FORCE BAUD RATE ON PRODUCTION UNITS!

### LINKED ISSUE #932: CoCo baud rate being forced with FORCE_UART_BAUD
FORCE_UART_BAUD is being added into the coco build files, which is causing production CoCo FujiNets to ignore chip select to select baud rate.

tschak909:
Big no no! It was causing coco units in the field to not do dip switch baud rate setting.


mozzwald:
This was already fixed in #934 for devkitc. The ESP32S3 boards are still prototype stage and will likely be a forced baud rate so no need to bother with these.

PR #935, opened 2025-06-02
Use constants for FujiNet commands instead of harcoded values.

FozzTexx:


tschak909:
Looking good, but I think apple2 is missing something. :)

FozzTexx:
> Looking good, but I think apple2 is missing something. :)

Check now

tschak909:
Awesome! Thank you!

tschak909 (approved):
moar readubiliti alwayz gud. :D :)

PR #943, opened 2025-06-08
Add v1.1.1 config-ng autorun

markjfisher:
Update config-ng autorun atr for alternate config from web ui.


markjfisher:
embedded version string is wrong. i'll redo this

PR #951, opened 2025-06-19
increase IEC buffers to 255 bytes

idolpx:
This should fix the issue with URLs getting cut off at 128 bytes.


tschak909 (approved):
Thank you.

PR #952, opened 2025-06-19
Make insert_boot_device into one method instead of two with ifdefs

FozzTexx:
I don't see why `#ifdef ESP_PLATFORM` is necessary since both versions are nearly identical. This seems a lot simpler to me, so I'm pretty sure I'm missing something.


tschak909:
Thank you.

PR #963, opened 2025-06-20
Simplify insert_boot_device() in iwm/fuji.cpp

FozzTexx:
Removed the `#ifdef ESP_PLATFORM` and simplified the method so that it does the same thing on both ESP32 and "PC".

Opened in draft mode, needs testing on both Apple II and FujiNet PC Apple II emulation.

image


tschak909 (approved):
Tested. Works for me.

PR #968, opened 2025-06-25
Fix PIO venv activate message output

benjamink:
Fixed the output message to be more accurate if the PIO `penv` has already been activated in the current environment.  

Activating venv (penv directory exists & isn't already activated): 
image

Activation failed with a bogus or missing PIO penv directory - script continues without it: 
image

`$VIRTUAL_ENV` is already set to the PIO penv so just continues without doing anything: 
image

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -160,6 +160,18 @@ if [ $BUILD_ALL -eq 1 ] ; then
 fi
 
 ##############################################################
+# Set up the virtual environment if it exists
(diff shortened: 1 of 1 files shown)


dillera:
Also please change the font in your images to Isoveka.

Screenshot 2025-06-25 at 4 38 27 PM

FozzTexx (approved):

PR #1005, opened 2025-08-31
Adds xdrive device support

fintros:
Hi Guys, 
What do you think about adding one more piece of hw support?
Please check this pull request and let ne know what do you think.
Some comments: 
1. Current XDrive version is continue of this device: https://forums.atariage.com/topic/371901-xdrive-yet-another-drive-emulator moved to ESP32S3 chip. And foirtunately now its able to boot and works with FujiNet fw.
2. Hardware repo is: https://github.com/fintros/xDrive - it's still not fully ready, need to create some readme and docs - will do it later. 
3. I've added standard esp-idf led_strip component, because I have addressable led on the device
4. Unfortunately should add fujinet_partitions_4MB_noupdate.csv because it seems after adding update partitin in the base file spiffs do not fit to the storage partition in 4Mb variant.

Waiting for your feedback,
BR, Alexey


dillera:
Is this for FN Firmware to be able to be flashed onto your xdrive hardware?

fintros:
Basically xdrive is the next version of sdrive2, so its originally designed for the SDrive firmware, but because it have esp32s3 on the board I've decided to check how FN fw will work on it . Fortunately everything works perfectly! Thanks for brilliant work!

mozzwald:
The build is failing with:

```/usr/bin/ld: CMakeFiles/fujinet.dir/lib/hardware/fnSystem.cpp.o: warning: relocation against `fnLedStrip' in read-only section `.text'
/usr/bin/ld: CMakeFiles/fujinet.dir/lib/hardware/fnSystem.cpp.o: in function `SystemManager::SystemManager()':
fnSystem.cpp:(.text+0x104): undefined reference to `fnLedStrip'
/usr/bin/ld: fnSystem.cpp:(.text+0x109): undefined reference to `LedStripManager::present()'```

**Comment by mozzwald** [comment]
> Sure will try to do it, but it seems this mechanizm is not used yet in the project - all other components included as source code.
> Does all of its modified from original one?

We have been testing the component manifest in a different branch for some other new drivers and it has worked. Something like this should do it I think, which goes in `src/idf_component.yml`:
```
## IDF Component Manager Manifest File
dependencies:
  espressif/led_strip:
    version: "3.0.1~1"
    rules:
      - if: 'target in ["esp32s3"]'
```

> Not tried it yet, but as I understand 4Mb Arari builds should fails right now because storage size is around 1.2Mb, but partition size defined as 768K... May be I miss some new configuration option?

Partition layout has changed recently with the addition of the update partition which is currently unused. I've never built 4MB firmware before so no idea if it even works. I was mostly curious if the flasher would work with this firmware file. To flash, you have to use the newest FujiNet-Flasher v1.5.0 since esp32s3 support was not available in the older versions. It is important that end users are able to flash firmware without having to use PlatformIO.

fintros:
Moved led_strip to the managed_components.
Fixed an issue with the build.
About FujiNet-Flasher v1.5.0 - proceed fast check, but it seems build.sh script generate firmware.zip with the empty "files" section in release.json file, will make more tests tomorrow.

fintros:
I think this variant should help.
I just moved #include to the same place where I define HAVE_LED_STRIP

FozzTexx (approved):

tschak909:
@fintros no, thank you. :)

PR #1007, opened 2025-09-05
add qrcode_minVersion() to auto size generated qrcode

idolpx:
minor tweaks to qrManager::to_petscii()

Now if you set version to "0" when calling QRManager::encode(), it will automatically select the minimum version to fit your data.


tschak909 (approved):
Thank you!

PR #1009, opened 2025-09-05
coco: fix debug/monitor baud rate

mozzwald:
This sets coco debug baud to match every other platform at 460800 baud


idolpx:
I've actually bumped things up to 2Mbps for monitor and upload.  Everything has been working fine.

PR #1013, opened 2025-09-07
Reworked the QRCode library

idolpx:
- If "version" is set to 0 it will automatically select the minimum version to fit the data.
- If encoding mode is not specified it will automatically select the best mode for the provided data.
- Added functions for generating QR Codes in ANSI and SVG formats for use in the serial console or web config.
- Fixed an issue that would not let you generate code versions 6 to 40.
- Eliminated a huge lookup table and replaced with a few calculations.
- Reduced memory usage overall.
- Modified the class interface to make it easier to use.

Minimal usage example
```c++
QRManager qrManager = QRManager(0, ECC_LOW, QR_OUTPUT_MODE_ANSI);
auto qrcode = qrManager.encode("This is some cool text for a QR Code!");
printf("%s\n", qrcode.data());
```


idolpx:
I added some more nullptr checks that may resolve the crash issue you had.

PR #1168, opened 2026-02-12
Adds TCPS protocol for TCP using TLS

arswaim:
Adds a new network protocol, TCPS, which allows for connections over TCP using TLS. It uses Mongoose to handle TLS, based very loosely on the HTTPS protocol.

I've successfully tested in- and out-bound connections from an Apple II, but I don't have access to other platforms FujiNet supports. I think it should work across platforms, but I've been wrong before.

Currently there is a self-signed cert directly in the code for when acting as a server. Clearly this is not a good solution, but I'm not sure what a good solution looks like here. Perhaps it should be possible to add a cert via the FujiNet's web interface? Also, I'm not sure how realistic a scenairo running a TCP server with TLS really is. It may be an edge case. 

Worse, the CA from `data/webui/common/ca.pem` is also in the header file, because I could not figure how to read it from the local file system. Loading a CA file can be skipped, but then there is no authentication. Any help with this would be greatly appreciated. I'm sure I'm missing something obvious.

I'm quite sure there are many problems with this code, and it may wind up being completely unusable. I'm happy to make any changes that are needed, and won't be offended if none of it passes muster.


tschak909:
Ok. I've tried to deal with the various issues of this PR.

But given that it has so many issues, I'm going to close it.

arswaim:
> Ok. I've tried to deal with the various issues of this PR.
> 
> But given that it has so many issues, I'm going to close it.

Sorry about that; I’ve been working on it locally, but have been distracted with other things. When it’s in better shape I’ll open a new PR.

PR #1193, opened 2026-02-26
[apple2] connect up webui config options

robjustice:
Hooked up for the Apple2
- config enable/disable button
- alternate config path to mount from SD card (as per webui help comment)

Added if config is disabled, and no alternate config path is set, then default to mount all. Not sure for this, have a look and see what you think. Needs to be later in the setup as its trying most likely to be mounting off tnfs server, so needs the wifi running.


tschak909 (commented):
Thank you! :)

tschak909 (approved):

PR #1238, opened 2026-04-08
[apple2] update to support 8 SP devices & updated config po

robjustice:
Update for Apple2 to support 8 Smartport devices.
And add updated supporting config autorun.po

It probably needs a bit more of a tidy up in regards to Fujinet-pc only supporting the 8 SP devs. That seems a little inconsistent with the conditions used.


FozzTexx:
Please make sure Disk II devices still work.

robjustice:
I have mounted a dos 3.3 disk in D1 and another in D2 and it boots ok. I can catalog the other one fine, and switch back. so looks ok from what i can see

tschak909:
This guy's got a conflict on autorun.po that needs resolving.

robjustice (commented):
hopefully this is ok

PR #1242, opened 2026-04-10
Fix intermittent wifi scan issues

RichStephens:
OK... I did this, because, at least on coco, with the latest firmware, I keep getting NO wifi networks found, and the log shows:

E (68595) wifi:2g bitmap contains only invalid channels= 0xffff

The change here affects all the other platforms, so I think some close scrutiny (and maybe testing) is necessary before merging this.  I can say this, it IMMEDIATELY fixed my wifi scan issues on coco.


tschak909:
Let's see how this shakes oout.

tschak909 (approved):

PR #1244, opened 2026-04-13
[apple2] update to support 8 SP devices & updated config image v2

robjustice:
Take2, let see if this works better.
Update for Apple2 to support 8 Smartport devices.
And add updated supporting config autorun.po


tschak909 (approved):
Thank you. :)

PR #1249, opened 2026-04-13
Lynx firmware cleanup and regresssion fix

shawnjefferson22:
Cleanup of commented out and also unused code.  Fixed regression for lynx fuji device for open directory (but probably more) where a fixed transfer of 256 was assumed in shared fuji device code, but Lynx Config sends variable length string.  transaction_get was failing if the read wasn't the fixed buffer size in length.


FozzTexx (approved):

tschak909:
Thank you!

PR #1258, opened 2026-04-18
fix iic rom0 issue with fn-lib-experimental

robjustice:
this removes a condition on the workaround previously added for the Liron. Its the same firmware bug in the IIc rom0. Looks like fn-lib-exp is sending the control cmds with the correct length and it triggers this problem to happen.


FozzTexx (approved):

tschak909 (commented):
Wow.

PR #1259, opened 2026-04-18
RC2014: start refactor to new internal API

juzzas:
There has been a lot of changes of internal API where the RC2014 port has been left behind.

This is the start of the refactoring process where we start by making the thing at least compile!

Note: Not all functionality has been tested (or wasn't event implemented in the first place), but simple fuji and network commands work.


tschak909 (approved):
Thank you good sir :)

tschak909 (approved):

FozzTexx:
You should add RC2014 to `.github/workflows/autobuild.yml` to keep it from getting behind again.

Another thing to consider is that if you're planning to switch to FEP-004/Fujiversal then the RC2014 stuff may not be needed at all in the firmware since then you'd use RS232.

FozzTexx (approved):

PR #1260, opened 2026-04-22
coco/msdos - implement full APETIME clock.  Atari: Fix APPLE3_SOS_BIN…

RichStephens:
…ARY format.

This has been tested with updates to fujinet-lib-experimental (PR for that to come), and a test program that tests every function AND EVERY FORMAT in fujinet-clock.h on Apple2, Atari, CoCo, and MS-DOS.


tschak909 (approved):
Looks good to me.

PR #1267, opened 2026-04-28
Coco: fix vfs file handles.  Prevent vfs file handle leaks.

RichStephens:


tschak909 (approved):
Looks good.

mozzwald:
fwiw, the sdkconfig.coco file was not previously tracked and probably does not need to be for this single vfs fix. Deleting your local sdkconfig.coco would force platformio to rebuild it automatically based on the updated defaults file that contained the vfs fix

FozzTexx:
> fwiw, the sdkconfig.coco file was not previously tracked and probably does not need to be for this single vfs fix. Deleting your local sdkconfig.coco would force platformio to rebuild it automatically based on the updated defaults file that contained the vfs fix

PR for adding the v1.5.2 fix to master: https://github.com/FujiNetWIFI/fujinet-firmware/pull/1270

PR #1271, opened 2026-04-28
Add Lynx Rev1 ESP32 S3 support

mozzwald:
* Added FujiNet Lynx Rev1 ESP32-S3 PlatformIO build support in platformio-sample.ini and new build-platforms/platformio-fujinet-lynx-rev1.ini
* Added Lynx Rev1 web UI config at data/webui/config/fujinet-lynx-rev1.yaml.
* Added new PINMAP_LYNX_S3 pin mapping with SD, UART, button, LED, and COMLYNX assignments ( the right way ;) )
* Enable/fix printf-based debug output for S3 based Lynx
* Added Lynx Rev1 hardware version detection/reporting
* Updated GPIO read/write helpers to use ESP-IDF gpio_get_level / gpio_set_level since S3 has more GPIO than older ESP32
* Increased card-detect task stack size so it doesn't crash
* Added Lynx S3 LED initialization for BUS, BT, and WIFI LEDs
* Remove bogus COMLYNX reset interrupt setup

### DIFF
--- a/.github/workflows/autobuild.yml
+++ b/.github/workflows/autobuild.yml
@@ -29,6 +29,7 @@ jobs:
           - IEC-LOLIN-D32
           - IEC-NUGGET
           - LYNX
+          - LYNX-S3
           - RS232
 
     steps:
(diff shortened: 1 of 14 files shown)


FozzTexx:
Add it to the workflow too so it doesn't get out of sync

FozzTexx (approved):

PR #1272, opened 2026-04-28
coco/pc: add chunked serial write throttle for USB-bridged downstream

RichStephens:
When fujinet-pc-COCO talks to a USB-tethered bridge cart (e.g. PicoROM), large response writes from drivewireNetwork::send_response can exceed the bridge's downstream buffer because tcdrain() on USB CDC-ACM only waits for USB ACK, not for the bridge to forward bytes to the CoCo bus.

Adds two optional [Serial] settings (PC builds only):
  chunk_size       - max bytes per ::write() syscall (default 0 = single write)
  chunk_delay_us   - usleep after tcdrain between chunks (default 0) - 500 works well for prototype bus cart 1.

Plumbed through TTYChannel::dataOut. Wired into drivewire bus setup. COMChannel ChannelConfig gets matching API stubs so Windows builds compile; behavior unchanged for builds that omit the settings.

### DIFF
--- a/lib/bus/drivewire/drivewire.cpp
+++ b/lib/bus/drivewire/drivewire.cpp
@@ -818,6 +818,10 @@ void systemBus::setup()
                       .baud(_drivewireBaud)
                       .deviceID(DW_UART_DEVICE)
                       .readTimeout(500)
+#ifndef ESP_PLATFORM
+                      .chunkSize(Config.get_serial_chunk_size())
+                      .chunkDelayUs(Config.get_serial_chunk_delay_us())
+#endif /* ! ESP_PLATFORM */
 #ifdef ESP_PLATFORM
                       .inverted(DW_UART_DEVICE == UART_NUM_2)
 #endif /* ESP_PLATFORM */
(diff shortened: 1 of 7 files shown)


FozzTexx:
I don't think adding sleeps is the right way to fix it. Plus adding fiddly config options to control it isn't right either.

This really needs to be fixed in the Fujiversal Pico firmware instead.

What are the steps to exhibit the problem?

RichStephens:
> I don't think adding sleeps is the right way to fix it. Plus adding fiddly config options to control it isn't right either.
> 
> This really needs to be fixed in the Fujiversal Pico firmware instead.
> 
> What are the steps to exhibit the problem?

Yes, you're probably right.  I wasn't really expecting this PR to survive scrutiny.  

To reproduce the issue:

All you have to do is fire up the protototype board connected to fujinet-pc-coco with the latest firmware, hit the lobby command from config to load the lobby. When the lobby program gets to the "Loading Servers..." message, the firmware (and the lobby program) hangs.

tschak909 (approved):

PR #1282, opened 2026-05-03
This fix problems were the base64 convertion was failing

pedgarcia:
when testing in Cubedot software in fujinet-pc (creating user/password tokens based on the base64 conversion)

The fix aligns Base64 commands with the same transaction_continue(NO_GET) sequencing already used by the Fuji hash commands in sioFuji.cpp.


tschak909:
awesome! Thanks!

FozzTexx (approved):

Issue #1119, opened 2026-01-09
ATARI Network (N:) device not reporting # of RX bytes in MASTER.


tschak909:
To reproduce:

* Load NOS from TNFS://APPS.IRATA.ONLINE/Atari_8-bit/DOS/NOS.atr
* CD N:TNFS://APPS.IRATA.ONLINE/
* DIR
* Directory contents should be displayed

The bad commit was traced to commit 065133cef4435ea78aa2f5dd073e56a0671c3a8b:

```
065133cef4435ea78aa2f5dd073e56a0671c3a8b is the first bad commit
commit 065133cef4435ea78aa2f5dd073e56a0671c3a8b
Author: FozzTexx <fozztexx@fozztexx.com>
Date:   Sun Nov 30 14:24:38 2025 -0800

    Remove rxBytesWaiting from NetworkStatus class since it's never current (#1095)

 lib/device/adamnet/network.cpp       | 47 ++++++++++++++++++-------------------------
 lib/device/comlynx/network.cpp       | 45 ++++++++++++++++++-----------------------
 lib/device/drivewire/network.cpp     | 88 +++++++++++---------------------------------------------------------------------
 lib/device/drivewire/network.h       |  5 -----
 lib/device/iec/network.cpp           | 34 +++++++++++++------------------
 lib/device/iwm/network.cpp           | 62 +++++++++++++++++++-------------------------------------
 lib/device/network.h                 |  9 ++++++++-
 lib/device/rs232/network.cpp         | 19 +++++++++---------
 lib/device/sio/network.cpp           | 43 +++++++++++++++++----------------------
 lib/fnjson/fnjson.cpp                | 13 ++++++------
 lib/network-protocol/FS.cpp          |  3 ---
 lib/network-protocol/FTP.cpp         |  1 -
 lib/network-protocol/HTTP.cpp        |  4 ----
 lib/network-protocol/Protocol.cpp    |  6 ++----
 lib/network-protocol/Protocol.h      |  2 ++
 lib/network-protocol/SSH.cpp         |  3 +--
 lib/network-protocol/SSH.h           |  2 +-
 lib/network-protocol/TCP.cpp         | 11 +++++++++-
 lib/network-protocol/TCP.h           |  1 +
 lib/network-protocol/Test.cpp        |  1 -
 lib/network-protocol/UDP.cpp         |  9 ++-------
 lib/network-protocol/networkStatus.h | 13 +-----------
[...]

PR #1286, opened 2026-05-12
remove SD::available() stub

a8jan:
looks like forgotten stub left in SD protocol, relates to https://github.com/FujiNetWIFI/fujinet-firmware/issues/1119
no one reading from SD? :-)


tschak909 (approved):
Ah. Thanks. :)

PR #1288, opened 2026-05-12
NetSIO, stop eating my CPU.

a8jan:
Added poll() back to netsio class (lost in refactoring).
Without poll(), SIO bus service() runs a busy loop, loading CPU on LWM and spinning fans.
Note: A similar change is needed for SIO serial port and probably Becker port too. I'll try to update them later.


tschak909 (commented):
Oh, THAT'S WHAT HAPPENED. 

Thank you for putting it back in! :)

FozzTexx (approved):

PR #1289, opened 2026-05-13
[atari] _sioBaud isn't the exact baud rate, just the state/mode

FozzTexx:


mozzwald (approved):
Modem baud change and high speed baud change work for me. 

850 relocator does a double load attempt and is slow/adds a delay. I don't recall if this was an issue in the previous baud fix attempt. I know this has been an issue in the past and has been fixed in the past.

PR #1292, opened 2026-05-13
Fix QROS scan loop causing SIO timeout on slow networks

viktorcech:
## Summary

The QROS detection loop in `check_for_FUJI_file()` had no offset limit and walked the entire CAS file looking for a `baud` chunk with `irg_length > 3000`. On standard 600 Bd CAS files (where the marker never exists) this caused hundreds of `fseek`/`fread` round-trips over TNFS — or extra iteration on top of the already-synchronous HTTP cache download — pushing total `mount all` time past the SIO command timeout, so CONFIG aborted.

This is the issue the DTIMLO 3× override workaround was added to mitigate. With this fix, the workaround should no longer be needed for typical cases.

## Fix

Limit the QROS scan to the first 8 KB. The QROS turbo `baud` chunk always sits near file start (right after the FUJI header, or after an optional ~2 KB 600 Bd boot block), so 8 KB has ~3× headroom and detection behaviour is unchanged for real QROS files. T2K scan was already limited to 256 B and is untouched.

```cpp
// before
while (scan_offset < filesize)

// after
const size_t QROS_SCAN_LIMIT = 8192;
while (scan_offset < filesize && scan_offset < QROS_SCAN_LIMIT)
```

One-line behavioural change plus an explanatory comment.

## Test plan

- [ ] Standard 600 Bd CAS over slow TNFS — should mount fast, no QROS misdetection
- [ ] T2K CAS — should detect as T2K, QROS scan never runs (`if (!tape_flags.turbo2000)`)
- [ ] QROS CAS at 6600 baud (e.g. EMOSOFT v5.1) — should detect and play
- [ ] QROS CAS at 9600 baud — should detect and play

### DIFF
--- a/lib/device/sio/cassette.cpp
+++ b/lib/device/sio/cassette.cpp
@@ -532,13 +532,15 @@ void sioCassette::check_for_FUJI_file()
         // QROS turbo uses 6580-6595 (AUDF=127) or 9535-9622 (AUDF=86).
         // Lower bauds (600, 854, 1000 etc.) are standard/KSO and handled
         // by the normal FUJI path.
-        // Scan limited to first 8 KB: QROS turbo baud chunk always appears
-        // near file start (after FUJI header and optional 600 Bd boot block,
-        // which is at most ~2-3 KB). Unbounded scan caused SIO timeouts on
-        // standard CAS files mounted over slow networks.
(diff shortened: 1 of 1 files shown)


tschak909 (approved):
This should be better.

PR #1297, opened 2026-05-16
Make SP-over-SLIP over USB generally work

oliverschmidt:
* The whole 'Bus Over Serial' (BOS) configuration settings are removed. The serial port name is determined by enumerating the exsisting serial ports and choosing the one that is a virtual port using USB. The classic serial connection parameters are meaningless for USB CDC-ACM.
* COMConnection::create_read_channel() is now structured like TCPConnection::create_read_channel().
* The COMConnection lifecycle is different from the TCPConnection one. I don't know if that is beneficial or a mere artifact. Anyhow, after a successful connection establishment and shutdown, no new connection can be established. Rather the process needs to be restarted. Hopefully somebody with a better understanding of the overall connection lifecycle management can improve that aspect.

### DIFF
--- a/fujinet_pc.cmake
+++ b/fujinet_pc.cmake
@@ -63,6 +63,7 @@ if(FUJINET_TARGET STREQUAL "APPLE")
         add_compile_definitions(SLIP_PROTOCOL_NET=1)
     elseif(SLIP_PROTOCOL STREQUAL "COM")
         add_compile_definitions(SLIP_PROTOCOL_COM=1)
+        set(USE_LIBSERIAL TRUE)
     endif()
 
     message(STATUS "SLIP_PROTOCOL is ${SLIP_PROTOCOL}")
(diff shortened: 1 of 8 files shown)


FozzTexx:
> Anyhow, after a successful connection establishment and shutdown, no new connection can be established.

That's why I used the BeckerSocket class in RS232/MSX. While it was originally intended just for CoCo, it's very smart about handling bringup and teardown of the connection and being able to re-establish a connection or wait for a new connection after being disconnected. It's also capable of both listening on a port to wait for a connection or acting as a client and connecting to a port.

I don't know if it would make sense for this use or not. The BeckerSocket class inherits from IOChannel and it works as a drop-in replacement for anything that is expecting to work with a "serial" port. If it would be useful here too, it might make sense to rename it and move the BeckerSocket class into a more generic location instead of continuing to reside in lib/bus/drivewire.

FozzTexx:
I guess I should also mention that conversion to BeckerSocket (if it even makes sense) can be done at a later time, not in this PR.

oliverschmidt:
@FozzTexx Thanks for you input! However, I believe that - just as you wrote - it is beneficial to have the functionality in question "somehow" available now.

FozzTexx (approved):

PR #1299, opened 2026-05-16
ACMChannel can now find the first USB CDC ACM serial device, no hardcoding of VID/PID

FozzTexx:


RichStephens (dismissed):
This smells like something to help my coco bus prototype :)

RichStephens (approved):

PR #1304, opened 2026-05-17
gcc preprocessor isn't as smart as it thinks it is.

FozzTexx:


tschak909 (approved):
tested and verified ok on ATARI.

PR #1306, opened 2026-05-18
Coco bus prototype: WiFi LED and activity (orange) LED "flicker" effect.

RichStephens:


FozzTexx:
I really don't like putting `#ifdef PINMAP_FUJIVERSAL_DRIVEWIRE` in things in lib/hardware. That stuff shouldn't need to know what board is being used.

RichStephens:
> I really don't like putting `#ifdef PINMAP_FUJIVERSAL_DRIVEWIRE` in things in lib/hardware. That stuff shouldn't need to know what board is being used.

Fair enough. I don' t know exactly how it's supposed to be organized, but I'll go back to the drawing board.

FozzTexx (approved):

PR #1310, opened 2026-05-20
Fix gdrive for fujinet pc

tschak909:
@FozzTexx  @a8jan  just wanted to close this loop, as routes and everything else needed to be set up for mgHttp and friends.

image

### DIFF
--- a/lib/http/mgHttpService.cpp
+++ b/lib/http/mgHttpService.cpp
@@ -10,9 +10,6 @@
 #include <ctime>
 #include <cstdlib>
 
-#include <cJSON.h>
-#include "mgHttpClient.h"
-
 #include "fnSystem.h"
 #include "fnConfig.h"
 #include "fnWiFi.h"
@@ -25,6 +22,8 @@
 #endif /* BUILD_ATARI */
 
 #include "mongoose.h"
+#include <cJSON.h>
+#include "mgHttpClient.h"
 #include "httpService.h"
 #include "httpServiceConfigurator.h"
 #include "httpServiceParser.h"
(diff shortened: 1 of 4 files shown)


lib/http/mgHttpService.cpp
```diff
@@ -679,6 +683,136 @@ int fnHttpService::get_handler_shorturl(mg_connection *c, mg_http_message *hm)
     return 0;
 }
 
+// ─── Google Drive OAuth2 relay handlers ──────────────────────────────────────
+
+#define GDRIVE_CLIENT_ID          "197927610161-me037pnh65lh9g8cad6fg62ifni9fik0.apps.googleusercontent.com"
```
FozzTexx, line 688: I'm a little concerned about having to put this in the source. I understand it isn't the key, but it still hardcodes in something that probably doesn't belong in the binary itself. Is there no better option?

FozzTexx (approved):

lib/http/mgHttpService.cpp
tschak909: I went back and forth on this, and this seems to be the best option.

FozzTexx: I wonder if maybe that string should be used as a default for fnConfig so that if necessary it could at least be overridden with fnconfig.ini. Still hardcoded into the source (ick) but wouldn't require rebuilding from source if for some reason it should be changed. (User wants to use their own ID somehow, current ID is no longer valid?)

a8jan:
@tschak909  Thank you for making gdrive available on FN-PC/LWM too!

PR #1311, opened 2026-05-20
Fix/atari ssh stability and startup mount idempotency

peterkaczorowski:
## Summary

  Three independent but related fixes discovered and verified while
  debugging SSH (`N:SSH://`) on a real Atari + real FujiNet hardware
  against a macOS-hosted TNFS server.  Each commit stands on its own and
  can be reverted independently, but together they make the SSH boot →
  connect → use → disconnect cycle stable for FujiNet.

  ### Commits

  1. **`8daea7e` — Make startup mount-all idempotent to avoid double-init race after WiFi async bring-up**

     `main_setup()` and the `IP_EVENT_STA_GOT_IP` handler in `fnWiFi.cpp`
     both ran the full TNFS mount-all sequence in parallel.  The second
     caller deleted and re-created the FileSystemTNFS underneath
     already-opened disk-image file handles, producing:

     failed seeking to header on disk image (-1, 9)
     read sector: error reading from disk image (...): -1

  (errno 9 = EBADF).

  Adds a tiny atomic guard `_startup_mount_lock` in `fujiDevice`
  plus a wrapper `fujicore_mount_all_at_startup()` used by both
  callers.  Whichever arrives first wins; the loser logs *"another
  caller already owns the startup mount, skipping"* and returns
  success.  On a **failed** attempt the lock is released so a later
  caller can retry (e.g. `main_setup` early → fails because WiFi has
  not yet acquired IP → releases → IP-GOT handler retries).  On a
  **successful** mount the lock stays held forever.

  User- and bus-initiated remounts intentionally **bypass** the
  guard so on-demand remount continues to work — only the startup
  path is gated.

  *A previous attempt to fix this with `!Config.get_wifi_enabled()`
  was correctly flagged by the FujiNet team for regressing two
  real-world scenarios (WiFi enabled + router out of range, Config
  disabled + WiFi enabled + no association).  The idempotent-guard
  approach keeps the original "always mount at startup" semantics
  intact for both scenarios.*

  2. **`5cb116b` — Fix heap corruption in NetworkProtocolSSH lifecycle and read path**

  Four defensive fixes that together stop the firmware crashing
  with *"heap_caps_free target pointer is outside heap areas"* on
  N: CLOSE after an SSH session:

  - **A.** NULL-initialise `session` and `channel` in `SSH.h` so
    close()/dtor are safe when `open()` failed early.
  - **B.** Tolerate failed `heap_caps_malloc(MALLOC_CAP_SPIRAM)` for
    `rxbuf` in both ctor and dtor.
  - **C.** Tear down `channel` before `session` in `close()`, with
    NULL guards, and chain to `NetworkProtocol::close()` like
    `TCP::close()` does (so the std::string buffers are properly
    cleared/shrunk).  Previously a remote "exit\n" left libssh
    channel state half-freed and the next N: CLOSE double-freed
    the session.
  - **D.** Filter `ssh_channel_read()` return value — accept only
    strictly positive byte counts.  `SSH_ERROR` (-1) was previously
    cast to `size_t` and fed to `std::string::append()`, which
    corrupted the PSRAM heap and only surfaced as the
    `heap_caps_free` abort later in the dtor.

  3. **`3f5c529` — libssh: stop assigning RODATA `""` to freed fields; NULL-check fopen path on ESP32**

  - **`options.c`:** `ssh_options_apply()` LIBSSH_FUJINET branch was
    assigning the RODATA literal `""` to `session->opts.knownhosts`
    and `session->opts.global_knownhosts`.  `ssh_free()` later
    `SAFE_FREE()`s those fields, calling `free()` on a pointer into
    read-only flash → `heap_caps_free` abort on every SSH close.
    Assigning `NULL` instead matches what all the known_hosts.c /
    knownhosts.c consumers already expect for "no file configured".

  - **`knownhosts.c`:** After the options.c fix above,
    `ssh_known_hosts_read_entries()` started receiving NULL
    filenames.  Plain libssh tolerates that because glibc's
    `fopen(NULL, "r")` returns NULL with `errno=EFAULT`, but **ESP32
    newlib's `fopen()` passes the path straight to `strlen()` inside
    `get_vfs_for_path()`** which dereferences it and panics:

    ```
    Guru Meditation Error: LoadProhibited
    PC: 0x40090815 (strlen+0x5 in ROM)  EXCVADDR: 0x00000000
    ```

    Treat NULL the same as "file does not exist": early return
    `SSH_OK` with no entries appended (matching the existing
    contract at the original lines 226-230).  `SSH_ERROR` was
    considered and rejected because it would propagate as
    `"Can't find a known_hosts file"` and break SSH connect for
    FujiNet, which has no persistent known_hosts file by design.

  ## How they relate

  All three were uncovered by the same end-to-end test (booting a real
  Atari, mounting boot disks over TNFS, then running `sshtest.com` /
  `ssh.com` against `N:SSH://`):

  - The startup mount race (#1) blocked boot from even completing
    reliably enough to test SSH.
  - The libssh RODATA bug (#3) crashed the firmware on the **first**
    N: CLOSE.
  - Fixing (#3) revealed that `ssh_known_hosts_read_entries()` was
    being reached on the connect path with a NULL filename — hence the
    second part of #3.
  - Fixing libssh exposed the latent heap-corruption in
    `NetworkProtocolSSH` (#2), which had been masked by the libssh
    crash happening first.

  ## Verification

  Tested end-to-end on real hardware (Atari 800XL + FujiNet ESP32 +
  macOS-hosted tnfsd) with:

  - Cold boot with WiFi enabled, router in range — boot mount succeeds
    via `main_setup`'s first attempt; IP-GOT handler logs the
    "skipping" message; no EBADF, no double mount.
  - Cold boot with WiFi enabled, router *not* in range — `main_setup`
    retries; eventually times out gracefully; SIO service loop runs
    normally (per FujiNet team feedback this scenario must always
    reach the bus loop).
  - `sshtest.com` open → banner → key-exchange → channel → write/read
    → exit cleanly; firmware logs the orderly teardown and remains
    responsive on N: for the next session.
  - Multiple back-to-back SSH connects from the same Atari boot session
    without firmware panic, heap-corruption warning, or `LoadProhibited`.

  ## Notes for reviewers

  - `_startup_mount_lock` is `std::atomic<bool>` with `exchange()` —
    no mutex, no allocations on the hot path; the wrapper itself
    cannot deadlock.
  - The libssh changes are inside the existing `LIBSSH_FUJINET`
    ifdef blocks, so upstream libssh consumers (glibc-based) are
    unaffected.
  - `knownhosts.c`'s NULL check is platform-agnostic and harmless on
    glibc — it just short-circuits the same failure path one frame
    earlier.
  - `NetworkProtocolSSH::close()` now chains to `NetworkProtocol::close()`,
    matching the pattern in `TCP::close()` and `UDP::close()`.

  ## Credit

  Bug #1 (startup mount race) regression scope was correctly identified
  by feedback from Mozwalld after the first attempted fix — the
  final idempotent-guard approach is the result of that review.

### DIFF
--- a/components/libssh/src/knownhosts.c
+++ b/components/libssh/src/knownhosts.c
@@ -222,6 +222,22 @@ static int ssh_known_hosts_read_entries(const char *match,
     FILE *fp;
     int rc;
 
+    /* Fix (2026-05-19, fujinet): the LIBSSH_FUJINET branch of
+     * ssh_options_apply() leaves session->opts.knownhosts and
+     * global_knownhosts as NULL (FujiNet has no persistent
+     * known_hosts file).  Callers (ssh_known_hosts_get_algorithms_names,
+     * ssh_session_has_known_hosts_entry, ...) pass that NULL straight
+     * through to here.  Plain libssh tolerates that because glibc's
+     * fopen(NULL, "r") returns NULL with errno=EFAULT, but ESP32 newlib's
+     * fopen() passes the NULL path to strlen() inside get_vfs_for_path()
+     * which then dereferences it and panics the firmware with
+     * "Guru Meditation Error: LoadProhibited" at strlen+0x5 in ROM.
+     * Treat a NULL filename the same as a missing file: success with
+     * no entries appended. */
+    if (filename == NULL) {
+        return SSH_OK;
+    }
+
     fp = fopen(filename, "r");
     if (fp == NULL) {
         SSH_LOG(SSH_LOG_WARN, "Failed to open the known_hosts file '%s': %s",
--- a/components/libssh/src/options.c
+++ b/components/libssh/src/options.c
@@ -1452,8 +1452,20 @@ int ssh_options_apply(ssh_session session) {
     free(session->opts.knownhosts);
     session->opts.knownhosts = tmp;
     #else
+    /*
+     * Fix (2026-05-19): the previous code assigned the RODATA literal
+     * "" to session->opts.knownhosts.  ssh_free() in session.c later
+     * calls SAFE_FREE(session->opts.knownhosts), which invokes free()
+     * on a non-heap pointer and aborts the firmware with
+     * "heap_caps_free target pointer is outside heap areas" on every
+     * SSH close.  All consumers in known_hosts.c / knownhosts.c
+     * already treat NULL as "no known_hosts file", and they emit the
+     * spurious log "Failed to open the known_hosts file '': Is a
+     * directory" when given an empty path, so NULL is both safer and
+     * cleaner than "".
+     */
     if (session->opts.knownhosts != NULL) free(session->opts.knownhosts);
-    session->opts.knownhosts = "";
+    session->opts.knownhosts = NULL;
     #endif
 
     #ifndef LIBSSH_FUJINET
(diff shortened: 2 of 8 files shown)


tschak909 (approved):
Thanks. We're going to trim down the commit message from a PhD dissertation. :)

PR #1315, opened 2026-05-23
Removed debug prints from ACMChannel

FozzTexx:
Removed debug prints from ACMChannel because they were slowing down being ready for data from Pico when CoCo boots.


RichStephens (approved):

tschak909 (commented):
Yeah, that makes sense. :)

PR #1317, opened 2026-05-25
Set term, rows, columns from devicespec url if passed-in for ssh and …

RichStephens:
…telnet


RichStephens:
Tested this with ssh to linux and telnet to a BBS.

FozzTexx (approved):

PR #1322, opened 2026-05-28
Atari fixes for alternate boot, config-ng, wrong filesize, and broken…

markjfisher:
… block directory reading

Restore Atari CONFIG-NG/alternate boot disk selection and fix directory block-mode crash introduced by the fujiDevice unification (#1108).

Boot disk selection was moved into generic fujiDevice::insert_boot_device() without the Atari-specific logic for SD alternate config paths and /autorun-cng.atr. Add an sioFuji override to restore the previous priority: alternate SD file, then CONFIG-NG, then default autorun.

Directory listings from CONFIG-NG crashed on READ DIRECTORY ENTRY in block mode (AUX2 $C0-$FF) because fujicmd_read_directory_entry() started an SIO transaction and then called fujicmd_read_directory_block(), which started a second one. Dispatch block mode before transaction_begin(), matching pre-unification behavior.

Also fix alternate-config SD mounts using FileSystem::filesize() instead of fsFlash.filesize() on handles opened from fnSDFAT.

### DIFF
--- a/lib/device/fujiDevice.cpp
+++ b/lib/device/fujiDevice.cpp
@@ -586,7 +586,7 @@ void fujiDevice::insert_boot_device(std::string boot_img, mediatype_t disk_type,
         return;
     }
 
-    image_size = fsFlash.filesize(fBoot);
+    image_size = FileSystem::filesize(fBoot);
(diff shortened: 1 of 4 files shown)


tschak909 (approved):
Thanks :)

PR #1323, opened 2026-05-28
Load ROM/ccc to pico on fujiversal coco prototype

RichStephens:
This is a DRAFT.  Do not merge yet, until we all agree on it.

## Summary

Adds a `MediaTypeROM` to the drivewire media layer so a `.rom`/`.ccc` file mounted on a device slot is streamed to the fujiversal PicoROM at mount time via the existing DBC `FujiBusPacket` protocol (`NETCMD_OPEN` / `NETCMD_WRITE` × N / `NETCMD_CLOSE`). Modeled on `rs232Disk::mountROM`.

All changes are gated `PINMAP_FUJIVERSAL_DRIVEWIRE` and only affect the `fujiversal-drivewire` build. No effect on rs232, other drivewire boards, or any other platform.

## What's added

- `MediaTypeROM` (`lib/media/drivewire/mediaTypeROM.{h,cpp}`): `mount()` streams the file to the pico in `MEDIA_BLOCK_SIZE` chunks; `read`/`write`/`format` return errors (this is a ROM, not a disk).
- `MEDIATYPE_ROM` enum entry, `.rom`/`.ccc` detection in `discover_mediatype()` (case-insensitive), wiring in `media.h` and the drivewire `disk.cpp` switch.
- Guards mount with `SYSTEM_BUS.isBoIP()` and returns an error under BoIP — the transfer is meaningless without a real pico.

## Drivewire `systemBus` plumbing

To send DBC packets over the bus, the drivewire `systemBus` gained `sendCommand` / `writeBusPacket` / `readBusPacket` mirroring the rs232 implementation.

Because DBC SLIP-framed packets and the CoCo's raw DriveWire bytes share the same USB-CDC link, `readBusPacket` captures any bytes seen before the SLIP frame start into a small pushback buffer (`_dbc_pushback`) instead of letting them be discarded. `_readByte()`, the public `read()`/`read(buf,len)`/`available()`, and `service()` all drain the pushback before consulting `_port`, so a stray CoCo opcode landing during a DBC read replays cleanly into the bus dispatch.

## Out of scope (follow-up)

Pico-side autostart command (drive `CART` low so the loaded ROM runs without the prototype's Q→CART jumper). Requires matching firmware on `~/fujiversal`.

## Testing

Built and flashed on `fujiversal-drivewire`. Verified end-to-end: mounted `.ccc` from the config program → ROM streamed to the pico → `fuji_mount_all()` returns → CoCo proceeds normally. With the Q→CART jumper installed on the prototype, the loaded ROM runs after reset.

### DIFF
--- a/lib/bus/drivewire/drivewire.cpp
+++ b/lib/bus/drivewire/drivewire.cpp
@@ -543,7 +543,7 @@ void systemBus::op_namedobj_mnt()
 // Read and process a command frame from DRIVEWIRE
 void systemBus::_drivewire_process_cmd()
 {
-    int c = _readByte();
+    int c = read();
     if (c < 0)
     {
         Debug_println("Failed to read cmd!");
--- a/lib/bus/drivewire/drivewire.cpp
+++ b/lib/bus/drivewire/drivewire.cpp
@@ -543,7 +543,7 @@ void systemBus::op_namedobj_mnt()
 // Read and process a command frame from DRIVEWIRE
 void systemBus::_drivewire_process_cmd()
 {
-    int c = _port->read();
+    int c = _readByte();
     if (c < 0)
     {
         Debug_println("Failed to read cmd!");
@@ -712,7 +712,7 @@ void systemBus::service()
         }
     }
 
-    if (_port->available())
+    if (available())
         _drivewire_process_cmd();
 }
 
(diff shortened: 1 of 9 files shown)


FozzTexx:
Probably don't have to `#ifdef PINMAP_FUJIVERSAL_DRIVEWIRE` in mediaType.cpp or disk.cpp. I can't see any reason to avoid detecting ROM media on the bitbanger cart. If anything, might want CONFIG to pop up and say "No, you need to upgrade your FujiNet!" 😉

Is SLIP_END 0xC0 already a Drivewire opcode?

FozzTexx:
If Drivewire isn't using 0xC0 as an opcode, would it not work to define OP_FUJIVERSAL as 0xC0 and then just process it in the switch like the others and avoid having to save and inspect every byte that passes through?

lib/bus/drivewire/drivewire.h
```diff
@@ -190,6 +194,21 @@ class systemBus
...
+#ifdef PINMAP_FUJIVERSAL_DRIVEWIRE
+        if (!_dbc_pushback.empty()) {
+            uint8_t b = _dbc_pushback.front();
+            _dbc_pushback.pop_front();
+            return b;
+        }
+#endif
+        return _port->read();
+    }
+
```
FozzTexx: This looks the same as the rewritten `read()` method below. Probably not needed?

lib/bus/drivewire/drivewire.cpp
```diff
@@ -543,7 +543,7 @@ void systemBus::op_namedobj_mnt()
 // Read and process a command frame from DRIVEWIRE
 void systemBus::_drivewire_process_cmd()
 {
-    int c = _port->read();
+    int c = _readByte();
```
FozzTexx: I think this can call the `systemBus::read()` just like you did with `available()` below.

lib/bus/drivewire/drivewire.h
```diff
@@ -331,13 +350,64 @@ class systemBus
...
+    size_t read() {
+#ifdef PINMAP_FUJIVERSAL_DRIVEWIRE
+        if (!_dbc_pushback.empty()) {
+            uint8_t b = _dbc_pushback.front();
+            _dbc_pushback.pop_front();
+            return b;
+        }
+#endif
+        return _port->read();
+    }
```
FozzTexx: To keep from having to handle _dbc_pushback in two places, the single byte read could call the multibyte read and fetch a single byte. That's how IOChannel works too.

FozzTexx (commented):
I suspect that all the `#ifdef PINMAP_FUJIVERSAL_DRIVEWIRE` isn't really needed. I don't think it will hurt anything if the bitbanger cart has the same routines that check for the SLIP packets.

I think the PR check failed because you need to add mediaTypeROM.cpp and mediaTypeROM.h to `fujinet_pc.cmake`

FozzTexx (approved):

PR #1325, opened 2026-06-02
AdamNet bus stability overhaul

tschak909:
Overhauls the AdamNet bus layer to fix the long-standing Coleco Adam reliability problems (random lockups, reboots, corruption). Develops in tranches; each commit builds for `fujinet-adam-v1` and was validated on hardware.

## Safety (the lockups/crashes)
- **Bound multi-byte receives.** `adamnet_recv()` busy-waited forever for the next byte; an aborted/garbled packet on the half-duplex wire spun the priority-17 bus task, starved IDLE1, and tripped the WDT into a reboot. It now times out and flags a frame error so handlers can abort.
- **Clamp wire-supplied lengths.** Device handlers fed the on-wire length straight into fixed buffers (serial/printer 16B, network 256B/1024B, fuji SSIDConfig) and `pkt_len--` could underflow to 0xFFFF. Every wire length is now clamped; the zero-length network packet is rejected.
- **Fix a null-deref crash in DDP/DSK high-score write-back** (`_media_host` was never assigned) and **never write back to the read-only config device** — a spurious write to autorun.ddp is consumed/ACKed but not applied.

## Half-duplex timing
AdamNet is one shared wire: every byte we transmit echoes into our own RX, and we must not drive while the master is mid-frame.
- **Deterministic echo drain.** Count the bytes we sent and consume exactly that echo, so a command the master has already started survives; glitch-tolerant (never blocks for a lost echo byte and swallows a real command).
- **Explicit minimum turnaround** replacing a 180us idle-discard that sat on the STATUS deadline; the response deadline is now measured after the turnaround.
- **Defer the Fuji WILL_GET ACK** until the whole frame is read — ACKing mid-frame ORed onto the inbound payload and corrupted directory paths.
- **Resync after a long blocking command** (WiFi scan, TNFS round-trips): such commands leave a backlog of the master's CONTROL.RECEIVE retries, so drain by idle detection instead of echo count.
- **ACK SET_DEVICE_FULLPATH** (it bypassed the transaction model and never ACKed, looping the mount).

## Result
From hard lockups to: clean boot + roll-call, reliable block I/O, WiFi scan, intact directory navigation, and disk image mounting that boots.

## Known limitation (out of scope, pre-existing)
Quick-boot (VI) from CONFIG stalls when the selected image is mounted onto D1 (the device the config disk runs from): EOS keeps D1's old DCB while the device now serves a different geometry. A hardware reset boots the selected disk correctly. This lives in the quick-boot / device-slot logic, untouched here.


FozzTexx:
This disk did a good job of testing the bus because it's CP/M and also does some file I/O, does network, and sends a few FujiNet commands. [lcltest.dsk.zip](https://github.com/user-attachments/files/28515918/lcltest.dsk.zip)

FozzTexx (approved):

tschak909:
done. Catharsis can begin.

PR #1331, opened 2026-06-08
[fujiDevice] use validate_host_slot function in fujicmd_copy_file_suc…

tschak909:
…cess

### DIFF
--- a/lib/device/fujiDevice.cpp
+++ b/lib/device/fujiDevice.cpp
@@ -966,15 +966,12 @@ success_is_true fujiDevice::fujicmd_copy_file_success(uint8_t sourceSlot, uint8_
         RETURN_ERROR_AS_FALSE();
     }
 
-    if (sourceSlot < 1 || sourceSlot > _totalDiskDevices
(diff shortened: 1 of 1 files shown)


FozzTexx (approved):
Approving, but are you sure about removing this
```
    sourceSlot--;
    destSlot--;
```

As I recall for whatever reason this command expects the indexing to be 1 based instead of 0 based like all the others.

FozzTexx:
> As I recall for whatever reason this command expects the indexing to be 1 based instead of 0 based like all the others.

@RichStephens is probably the expert on the copy command arguments since he was the one that has most recently fought with getting copy working in CONFIG.

RichStephens:
> > As I recall for whatever reason this command expects the indexing to be 1 based instead of 0 based like all the others.
> 
> @RichStephens is probably the expert on the copy command arguments since he was the one that has most recently fought with getting copy working in CONFIG.

That is absolutely true.  When fuji_copy_file() is called in the config, it adds 1 to source and destination slots numbers to account for that.

PR #1336, opened 2026-06-10
Fix crash when trying to do FUJICMD_RANDOM_NUMBER

### LINKED ISSUE #1335: Crash in FUJICMD_RANDOM_NUMBER
0x400f02ad: sioFuji::transaction_put(...) at lib/device/sio/sioFuji.h:51
  0x400f0b9d: sioFuji::sio_random_number() at lib/device/sio/sioFuji.cpp:944
  0x400f22fe: sioFuji::sio_process(...) at lib/device/sio/sioFuji.cpp:1303
  0x400dcbaa: systemBus::_sio_process_cmd() at lib/bus/sio/sio.cpp:305
  0x400dccf1: systemBus::service() at lib/bus/sio/sio.cpp:441
  0x40144b7f: fn_service_loop(void*) at main.cpp:578

  The crashing command is this frame:

  CF: 70 d3 20 00 64

  That is FujiNet device $70, command $D3, which maps to FUJICMD_RANDOM_NUMBER.

  Root cause: lib/device/sio/sioFuji.cpp:941 has:

  void sioFuji::sio_random_number()
  {
      int r = rand();
      transaction_put(&r,sizeof(int),true);
  }

It calls transaction_put() without first calling transaction_begin(TRANS_STATE::NO_GET), so the assertion in lib/device/sio/sioFuji.h:50 fires because _transaction_state is still INVALID.

There’s also a likely second bug there: it passes true as the err argument to transaction_put(), which means the random-number response is sent as an error. The fix should be:

  void sioFuji::sio_random_number()
  {
      int r = rand();
      transaction_begin(TRANS_STATE::NO_GET);
      transaction_put(&r, sizeof(r), false);
  }

FozzTexx:
Fixes #1335 

[lcltest.atr.zip](https://github.com/user-attachments/files/28797044/lcltest.atr.zip)

To test: run the attached disk image on Atari with current firmware, choose CRYPTO.TST. Flash to new firmware build by this PR under Checks, run CRYPTO.TST again.

Before fix:

image

After fix:

image

### DIFF
--- a/lib/device/sio/sioFuji.cpp
+++ b/lib/device/sio/sioFuji.cpp
@@ -843,8 +843,9 @@ void sioFuji::sio_base64_encode_output()
 
 void sioFuji::sio_random_number()
 {
+    transaction_begin(TRANS_STATE::NO_GET);
     int r = rand();
-    transaction_put(&r,sizeof(int),true);
+    transaction_put(&r,sizeof(int),false);
 }
 
 void sioFuji::sio_base64_decode_input()


tschak909 (approved):
Thanks.

PR #1339, opened 2026-06-11
[adam] oops.

tschak909:
Due to a stupid mistake, the config.ddp had to be re-packaged. I'm batting zero today.


FozzTexx (approved):

markjfisher:
Is it coincidence that "oops" is an anagram of "poos"? Maybe it's trying to tell you to use better commit messages.

Just a thought.

PR #1343, opened 2026-06-12
[apple2] fix cpm protocol task starting

robjustice:


tschak909 (approved):
Ok, will test this, and potentially drop the guard later. Thanks.

PR #1349, opened 2026-06-15
Fix clock_get_tz so that it returns UTC if no timezone is specified i…

RichStephens:
…n config.


FozzTexx (approved):
I do think the clock classes should work without this change too, but that should probably wait until all the clock classes are 1108.

PR #1350, opened 2026-06-16
atari: Add option to boot with High Speed SIO for disks, fuji and network

mozzwald:
This adds a new high speed SIO option to config for Atari. This option is not persistent and must be enabled in config each time by pressing `H`. The reasoning for this is some apps/games will not work with it and the user should explicitly enable this option only for disks they know will work. This should help prevent "nothing works" failures due to high speed problems. 

When enabled, the new HISIO disk is mounted instead of the users selected disks when they press OPTION to boot. The HISIO disk loads the patched SIO routines into RAM, mounts the users disks and continues to boot from D1. BOOT ERROR always appears once after the switch to booting the users D1 disk. This is normal.

This also requires patches to config which will be a separate PR there. The HISIO handler routines and disk creation are at [https://github.com/mozzwald/highspeed-sio/tree/fujinet](https://github.com/mozzwald/highspeed-sio/tree/fujinet). I'm opening this as a draft in the hopes that others can test this for any issues before it being accepted. I would really appreciate if @tschak909 could test this.


tschak909 (approved):
yay

mozzwald:
> yay

lol, too soon :laughing:

tschak909:
| Result | Desc                                                                    |
|--------|-------------------------------------------------------------------------|
| PASS   | Jumpman.atr from fujinet.online (HSIO Index 9)                          |
| PASS   | Abracadabra.xex from irata.online                                       |
| PASS   | Booting Pinball Construction set.atx from irata.online (it shifts down) |
| PASS   | Booting Chompelo.atr from irata.online                                  |
| PASS   | Pressing RESET boots original unpatched OS.                             |
| PASS   | Switch to HSIO Index 0, reboot Chompelo.atr                             |

Basically:

* The Idea is sound.
* It works for the most part.
* ATX packaged software will ignore high speed by design.
* Some ATR cracked software will fail.
* XEX works, and works extremely well.

I'd say, yeah, let's fold this in there.

PR #1351, opened 2026-06-16
Update dragon support

wdathing:
Dragon support got broken in implementation of new communications architecture.  Dragon is weird in that it's UART requires it to be only half inverted.
Also updated config apps for dragon to be new fujinet-lib based version.
An binary of lobby for dragon is also provided, but can't get to it from config (yet).  If you want to try it, do DWLOAD "DGNLOBBY.DWL" from basic.
loader/splash screen is in progress.


lib/hardware/ESP32UARTChannel.cpp
```diff
@@ -51,9 +53,15 @@ void ESP32UARTChannel::begin(const ChannelConfig& conf)
...
-
-    if (conf.isInverted)
-        uart_set_line_inverse(_uart_num, UART_SIGNAL_TXD_INV | UART_SIGNAL_RXD_INV);
+    
+
+    uint32_t inv_mask = 0;
+    if (conf.isInverted || conf.isTxInverted)
+        inv_mask |= UART_SIGNAL_TXD_INV;
+    if (conf.isInverted || conf.isRxInverted)
+        inv_mask |= UART_SIGNAL_RXD_INV;
```
FozzTexx, line 59: Makes more sense to me to get rid of `conf.isInverted` field completely and keep only `conf.isTXInverted` and `conf.isRXInverted`. In `ESP32UARTChannel.h` the `inverted(bool inv)` can do what it's doing now and setting the individual RX/TX flags and not bother with the now obsolete `isInverted`.

lib/hardware/TTYChannel.h
```diff
@@ -10,6 +10,9 @@ struct ChannelConfig
 {
     std::string device;
     int baud_rate;
+    bool isInverted = false;
```
FozzTexx: Same here and `COMChannel`, I don't see a reason to keep the separate `isInverted` when the new `isRXInverted` and `isTXInverted` combine to do the same thing.

lib/bus/drivewire/drivewire.cpp
```diff
@@ -812,7 +814,11 @@ void systemBus::setup()
     _drivewireBaud = FORCE_UART_BAUD;
   #else /* !FORCE_UART_BAUD */
     _drivewireBaud = readBaudSwitch();
-  #endif /* FORCE_UART_BAUD */
+    if (bDragon)
+    {
+        // Additional configuration for Dragon mode if needed
+    }
```
FozzTexx: This isn't doing anything here, I think it can be removed.

FozzTexx (changes requested):

lib/hardware/TTYChannel.h
wdathing: I did it this way to minimize the chance of breaking the platforms i can't test... i believe adam platform was inverting the uart, but i will go ahead and change this.

FozzTexx: > I did it this way to minimize the chance of breaking the platforms i can't test... i believe adam platform was inverting the uart, but i will go ahead and change this.

The field itself is redundant after adding the isTX/isRX fields. I want to keep the `inverted()` setter that still sets both at once.

I can test on Adam.

lib/hardware/ESP32UARTChannel.cpp
wdathing: Done

lib/hardware/TTYChannel.h
wdathing: Done

lib/bus/drivewire/drivewire.cpp
wdathing: Done

lib/hardware/COMChannel.h
```diff
@@ -22,6 +24,17 @@ struct ChannelConfig
...
+        isRxInverted = inv;
+        isTxInverted = inv;
+        return *this;
+    }
+    ChannelConfig& rxInverted(bool inv) {
+        isRxInverted = inv; return *this;
+    }
+    ChannelConfig& txInverted(bool inv) {
+        isTxInverted = inv; return *this;
+    }
```
FozzTexx: There is no way to invert the signal on COMChannel or TTYChannel, the addition of inverting shouldn't be here.

lib/hardware/ESP32UARTChannel.cpp
```diff
@@ -43,6 +43,8 @@ void ESP32UARTChannel::begin(const ChannelConfig& conf)
     case 2:
         rx = PIN_UART2_RX;
         tx = PIN_UART2_TX;
+        //gpio_set_pull_mode((gpio_num_t)rx,GPIO_PULLUP_ONLY);
+        //gpio_set_drive_capability((gpio_num_t)PIN_UART2_TX, GPIO_DRIVE_CAP_3);
```
FozzTexx: Since this isn't being used it shouldn't be added.

FozzTexx (changes requested):

lib/hardware/ESP32UARTChannel.cpp
wdathing: Done

wdathing:
General question: at $dayjob, our protocol is that the  opener of question is the one to hit the resolve button... same here, or when i make the fix do i hit resolve.
Thanks, and thanks for the feedback!

FozzTexx:
> General question: at $dayjob, our protocol is that the opener of question is the one to hit the resolve button... same here, or when i make the fix do i hit resolve.

My preference is the same as your $dayjob, but everyone handles it differently here.

FozzTexx (approved):
Adam boots, Atari boots, CoCo boots.

Looks good to me!

PR #1352, opened 2026-06-16
Fix off-by-one error in file copy

### LINKED ISSUE #1345: Copy file broken in current firmware
Have tested with CoCo firmware v1.6.0. Copy works. Copy does not work in current nightly.

It very likely is broken for all platforms.

The copy command needs to switch back to 1-based indexing for the host slots to maintain backward compatibility and all platforms tested for the copy command.

RichStephens:


RichStephens:
This PR fixes firmware issue: https://github.com/FujiNetWIFI/fujinet-firmware/issues/1345

tschak909 (approved):
ok

PR #1359, opened 2026-06-19
MSX ROM types using ROM DB

nwah:
DRAFT - need to settle on how/where to store the ROM database; how to only do this for MSX (?); if it's fine to read in the whole ROM up front, or if not how/when to send that over to the DBC

- adds `romType.h` with ROM type definitions
- adds `RomIdentifier` class that takes the hash of a ROM file and looks it up in a ROM user-updatable database file (`SD/msxromdb.txt`)
- updates `rs232/disk.cpp`'s `mountROM()` method to first read in the full ROM data, run it through `RomIdentifier`, and then pass the ROM type to the DBC during the `OPEN` call


FozzTexx:
I think your editor went cuckoo on CMakeLists.txt and fujinet_pc.cmake. It made a ton of diffs that aren't necessary.

My first impression is I think we need to 1108 lib/media sooner rather than later. Having all the media types separated by platform doesn't make sense and we're quickly moving into media types that can be used by more than one platform.

We also need to figure out a better way to deal with ROM media instead of pretending it's a disk. A problem that has already shown up on CoCo where it has both ROM *and* disk is that mounting a ROM makes it also appear in a disk drive! What happens if a user selects more than one ROM?

I feel like msxromdb.txt may not belong in the repo, but I don't know where else to put it. Is it getting written to the flash file system?

What do you mean by "if it's fine to read in the whole ROM up front"? You have to send the whole ROM over to the DBC before making it active.

PR #1361, opened 2026-06-19
Fix/util long entry size precision

RichStephens:
This is a small tweak to get slightly more accurate filesize in the directory listings.


FozzTexx (changes requested):
This seems to be a mix of other PRs and not just the stated changes.

PR #1363, opened 2026-06-19
Add 16/24/32 bit unsigned types that store their bytes in a specific endianness

FozzTexx:
Will be using this with some upcoming changes to Apple II, but I think it will be useful for all "over the bus" data on all platforms. These types can be inserted into a packed struct and can be used where normal unsigned types would work, such as in math or passing into larger integer arguments. They take up the exact number of bytes specified (2, 3, or 4) and are also stored in the struct with matching endianness so no extra steps are required to read/write the values before using or sending to the retro computer.

### DIFF
--- a/include/fuji_endian.h
+++ b/include/fuji_endian.h
@@ -13,7 +13,7 @@
 
 #endif
 
-#if defined(__linux__) || defined(__CYGWIN__)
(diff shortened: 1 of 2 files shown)


tschak909 (approved):
yup ok.

PR #1368, opened 2026-06-20
[iwm] Use parameter names in iwm_decoded_cmd_t instead of uint8_t array

FozzTexx:
Removed the `uint8_t` arrays in `iwm_decoded_t` using hardcoded indexes and replaced them with explicit parameter names for each command field, grouped by command type. This improves readability and makes the structure more self-documenting, making it clear which fields are being examined and what type of command packet is expected.

Also replaced the 16-bit and 24-bit fields in `iwm_decoded_t` with `u16le_t` and `u24le_t` types so that values can be accessed like regular unsigned integers, with no need to call additional methods to handle decoding.


tschak909 (approved):
Ok cool

PR #1369, opened 2026-06-20
[iwm] Fix HTTP POST

FozzTexx:
`set_channel_mode()` was getting the wrong values. Removed several no longer used variables which were confusing things.

Before:

image

After:

image


tschak909 (approved):
awesome.

PR #1373, opened 2026-06-22
[iwm] Use spError_t constants whenever an spError_t is used.

FozzTexx:
Changed `spError_t` to an `enum class` so the compiler would enforce that the correct error types were being used in the right places.


tschak909 (approved):
Cool

PR #1384, opened 2026-06-25
[adam] remove lib/compat/pc_rtos

FozzTexx:
Didn't need the lib/compat/pc_rtos that Claude added. Printer has been tested and it still works.

Opened against the PR 1379 branch adam-pc-boip-transport, not master, so it can be included in the PR.


tschak909 (approved):
Cool

PR #1390, opened 2026-06-27
netstream: fix unbounded downstream pacing + bufferbloat (frame-align…

brad-colbert:
…ed drop, configurable depth)

Fixes growing/unbounded downstream (server->Atari) latency in the concurrent UDP netstream SIO path (Mode B, Fuji cmd 0xF0 / FUJICMD_ENABLE_UDPSTREAM).

Pacing (eliminates the runaway):
- pace_to_atari() inter-byte gap now tracks the negotiated baud (10,000,000 / netstream_baud, 8N1) instead of a fixed 520us calibrated for 19.2k while the lane negotiates ~31960. The NetSIO hub credit flow-control remains the hard backstop, so the Atari can't be overrun.
- All queued UDP datagrams are drained per service pass (not one), including inside the outbound Atari->Net loop, so inbound intake isn't starved during bidirectional traffic. Per-call drain cap raised (NETSTREAM_PACE_MAX_PER_CALL).

Bufferbloat (keeps standing latency small at high rate):
- For the lossy UDP path, inbound is a bounded queue of whole datagrams/frames with drop-oldest eviction. enqueue_frame() tracks per-frame lengths (frame_len[] descriptor ring) and evicts the OLDEST WHOLE frame(s) over the depth cap. The frame currently being sent is copied into inflight[], outside the droppable queue, so eviction is always whole-frame aligned and never splits a frame mid-send. This matters because the Atari deframer pulls fixed-size chunks with no resync marker -- a byte-misaligned drop would desync it permanently. MIDI (port 5004) and TCP keep the lossless byte path (enqueue_rx, full ring, drop only on true overflow).

Configurable depth (#define default + ini override):
- NETSTREAM_RX_MAX_FRAMES (default 4, min effective 2) is the shallow-buffer depth in whole frames. New [Network] netstream_rx_depth in fnconfig.ini overrides it; 0 means "use the firmware default". Clamped to [2, slots-1].

Footprint (ESP32 DRAM is tight):
- inflight[] is sized to NETSTREAM_MAX_FRAME (256) rather than the full 2KB buffer; the framed path only carries small real-time datagrams, and larger ones are dropped whole. Keeps .dram0.bss within the classic ESP32 limit.

Verified end-to-end via the seq-echo --hz sweep on the PC build: hz=30 ~2.5s -> ~0.5s, hz=60 ~1.5s -> ~0.43s; ovf=0, bad=0, stale=0 and resync flat at every rate (whole-frame drops, no desync), confirmed down to rx_depth=2. A ~49.7k baud run confirmed raw bandwidth was never the limiter. Builds clean for both fujinet-pc (ATARI) and ESP32 (fujinet-atari-v1; RAM 36.9%, Flash 25%).

### DIFF
--- a/lib/config/fnConfig.h
+++ b/lib/config/fnConfig.h
@@ -118,6 +118,7 @@ public:
     int get_network_netstream_port() { return _network.netstream_port; };
     bool get_network_netstream_register() { return _network.netstream_register; };
     int get_network_netstream_mode() { return _network.netstream_mode; };
+    int get_network_netstream_rx_depth() { return _network.netstream_rx_depth; };
     bool get_general_config_enabled() { return _general.config_enabled; };
     void store_general_devicename(const char *devicename);
     void store_general_hsioindex(int hsio_index);
@@ -134,6 +135,7 @@ public:
     void store_netstream_port(int port);
     void store_netstream_register(bool enable);
     void store_netstream_mode(int mode);
+    void store_netstream_rx_depth(int depth);
     bool get_general_fnconfig_spifs() { return _general.fnconfig_spifs; };
     void store_general_fnconfig_spifs(bool fnconfig_spifs);
     bool get_general_status_wait_enabled() { return _general.status_wait_enabled; }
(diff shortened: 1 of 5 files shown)


mozzwald:
Tested with MidiMaze and it works still

mozzwald (approved):
Seems legit 😉

PR #1371, opened 2026-06-21
RunCPM 6.9: single shared core, N:CPM telnet console (:8677), RSX bar commands

peterkaczorowski:
## Summary

- **RunCPM 6.9 refactoring** — consolidate the three duplicate RunCPM engines (SIO `'G'`, `N:CPM://`, telnet) into ONE shared core. `runcpm_core.cpp` is now the only TU that compiles the Z80/CP/M chain (external linkage); the transports are thin console back-ends that plug into it through `runcpm_session.h`'s `runcpm_console_ops`, guarded by a `g_busy` single-instance interlock. `abstraction_fujinet_core.h` holds the shared disk/SD section. `preTables` moves the ~11 KB Z80 flag tables to flash and the consolidation reclaims ~22 KB, fixing the ESP32 `dram0_0_seg` overflow.
- **N:CPM telnet console on port 8677** — new `vm_telnet.cpp` exposes a loopback TELNET server (`127.0.0.1:8677`) bridged through libtelnet to the shared core, so an Atari reaches CP/M via `N:TELNET://127.0.0.1:8677`. It is started from `fn_service_loop` only after lwIP is up (`esp_netif_init` inside `fnWiFi.start`); the acceptor waits for an IP and is pinned to core 0 so the priority-17 SIO loop on core 1 cannot starve it.
- **RSX bar commands** — new `vm_bar.cpp/.h` implement the VM-generic bar-command gate (`|ftp`, `|wget`, `|fn`, `|apt`, `|help`) wired into `ccp.h` via `vm_bar_io` trampolines built from the core's `_getch/_putch/_puts`, so the bar talks to whichever transport is active. Named `vm_` (not `cpm_`) so future non-CP/M VMs can reuse the same telnet console + bar plumbing.

## ESP32 hardware bring-up fixes

Found and fixed on real Atari FujiNet (fujinet-atari-v1) hardware:
- **Boot loop** (`assert tcpip_send_msg_wait_sem ... Invalid mbox`): the telnet acceptor opened a socket before lwIP existed. Moved `vm_telnet_start()` out of `main_setup` into `fn_service_loop`, after the WiFi/BT block, gated on the network stack being up.
- **Listen-before-IP**: acceptor now waits for `fnWiFi.connected()` before `bind()/listen()`.
- **"Port open but no CP/M banner"**: the priority-10 acceptor pinned to core 1 was permanently starved by the priority-17 SIO loop. Pinned the acceptor to core 0.

## Test plan

[...]

PR #1391, opened 2026-06-27
RunCPM: upgrade to 6.9 and consolidate transports onto one shared core

peterkaczorowski:
## Summary

Updates the vendored RunCPM engine to **6.9** and refactors all CP/M transports onto **one shared stateless core**.

The 6.9 bump and the core refactor are in the same PR **by necessity**: 6.9 has substantially more `static` global state than 6.5, and each CP/M transport TU (`siocpm`, `drivewire`, `iwm`, `rc2014`, `rs232`) carries its own private copy via internal linkage. Five copies of 6.9's larger state overflow the ESP32 DRAM `.dram0.bss` segment by ~58 KB and the firmware will not link. Consolidating to a single engine instance is the prerequisite that makes 6.9 fit.

### What changed
- `runcpm_core.cpp` is now the only TU that compiles the Z80/CP/M chain (external linkage). `runcpm_session.h` exposes `runcpm_console_ops`; `abstraction_fujinet_core.h` holds the shared disk/SD section.
- The transports (`siocpm`, `drivewire/cpm`, `iwm/cpm`, `rc2014cpm`, `rs232cpm`, `network-protocol/CPM`) become thin console back-ends that plug into the shared core through `runcpm_console_ops`, guarded by a single-instance interlock.
- A read-handle cache collapses RunCPM's per-128-byte-record open/seek/read/close on the SD card into one cached read-only handle shared by `_sys_readseq`/`_sys_readrand`.
- Symbol-collision renames so the core TU can pull in bus headers on Apple/COCO targets: RunCPM `Command` struct → `CcpCommand` (collides with FujiNet `class Command`); `CPM` macro → handled in `globals.h` banner.

## Test plan
- [x] ESP32 `fujinet-atari-v1`: `./build.sh -b` links clean — **RAM 36.8%**, no `dram0_0_seg` overflow (vs. overflow-by-58 KB without the consolidation).
- [ ] SIO `'G'` CP/M and `N:CPM://` regression on hardware.

---
This is **PR 1 of 3** of a dependency stack that replaces #1371. Merge first; PRs 2 and 3 build on it.


peterkaczorowski:
Thanks for the review @FozzTexx — both points addressed.

**"Touching a lot more than just upgrading runcpm"**

It unfortunately can't be a pure version bump. 6.9's larger static state, replicated across all five transport TUs (SIO, IWM, DriveWire, RC2014, RS232), overflows the ESP32 `.dram0.bss` segment by ~58 KB — it won't link. The fix is to stop compiling RunCPM five times: the transports now share a single stateless core (`runcpm_core.cpp` + `runcpm_session.h`) and each bus just provides a small console back-end. So the 6.9 upgrade and the shared-core consolidation are one inseparable change — you can't land 6.9 without it.

To make that reviewable I split the PR into two commits:

1. **`Update vendored RunCPM to 6.9 (verbatim upstream)`** — a clean drop of the 10 stock upstream files, byte-for-byte identical to MockbaTheBorg/RunCPM @ 6.9. You can review this one in isolation by diffing against upstream; it should be empty.
2. **`FujiNet: integrate RunCPM 6.9 onto a single shared core`** — all the FujiNet-specific work: the shared core, the linkage shim, the name-collision renames, and the transport back-ends. This is the commit to actually review.

**"Walls of comments"**

Stripped them throughout — the integration commit is now net **-1137 lines**. I also deleted three abstraction headers that the shared-core refactor made dead code.

Build stays green at 36.8% RAM (`fujinet-atari-v1`). The two stacked PRs (#1392 telnet console, #1393 bar commands) are rebased on top of this.

peterkaczorowski:
Superseded by a cleaner split: #1401 (consolidate bus CP/M devices onto a shared cpmDevice base, on the current 5.8 engine) followed by #1402 (engine bump 5.8 → 6.9 on that base). Closing in favour of those.

PR #1401, opened 2026-06-29
RunCPM: consolidate bus CP/M devices onto a shared cpmDevice base class

peterkaczorowski:
## Summary

Introduces a single shared `cpmDevice` base class under `lib/device/cpm/` and
re-points every bus CP/M device at it, plus a single shared RunCPM engine
translation unit. This is **pure consolidation against the current RunCPM
(5.8)** — no engine version bump — addressing FozzTexx's request that the bus
CP/M devices inherit from one base class to handle only their minor differences,
so CP/M can eventually be enabled on all platforms.

## What changed

- **New `lib/device/cpm/cpm.{h,cpp}`** — `cpmDevice : virtualDevice` with the
  CP/M lifecycle (`cpmActive`, `init_cpm()`, `handle_cpm()`) and four virtual
  endpoint primitives (`ep_kbhit`/`ep_getch`/`ep_putch`/`ep_clrscr`) bridged to
  the engine's C callbacks via a static active-instance trampoline. A default
  endpoint cleanly exits the session for transports without a real console.
  `cpmQueueDevice` adds the byte-queue model used by the background-task
  transports (iwm/drivewire).
- **New `lib/runcpm/runcpm_core.cpp` + `runcpm_session.h`** — one TU compiles the
  Z80/CP/M engine with global linkage; transports call `runcpm_session_run()`
  with console callbacks and `runcpm_session_request_exit()` to stop.
- **Bus devices** (sio/rc2014/rs232/drivewire/iwm) now derive from the base and
  implement only their endpoint specifics. `N:CPM://` (network-protocol/CPM.cpp)
  folded onto the same core.
- **Removed** the duplicated `abstraction_fujinet_apple2.h` and
  `abstraction_network_protocol.h` (−1.8k lines); `abstraction_fujinet.h`
  slimmed to delegate to the shared core.

Net: **+613 / −1843** across 25 files.

## Build / test

- Desktop (`fujinet_pc.cmake`): ATARI, APPLE, RS232 targets link clean.
- ESP (PlatformIO/ESP-IDF): `fujinet-atari-v1`, Apple, CoCo build clean.
- Verified on real FujiNet ESP32 (Atari): RunCPM over SIO `G`/R: and
  `N:CPM://` both run; bus CP/M sessions start/exit correctly.

## Notes

- Stays on RunCPM 5.8 deliberately so this is reviewable as a structural change,
  not an engine upgrade. The 6.9 bump rides on top of this base in a follow-up.
- Transports that don't support an interactive console get the clean-exit
  endpoint stub rather than hardcoded values.


lib/device/iwm/cpm.cpp
FozzTexx, line 94: Shouldn't this be using the new available() method?

lib/device/drivewire/cpm.cpp
```diff
@@ -80,42 +37,34 @@ void drivewireCPM::send_response()
...
-#ifdef ESP_PLATFORM
     if (cpmTaskHandle != NULL)
     {
         vTaskDelete(cpmTaskHandle);
         cpmTaskHandle = NULL;
     }
 
     xTaskCreatePinnedToCore(cpmTask, "cpmtask", 32768, NULL, 20, &cpmTaskHandle, 1);
-#endif /* ESP_PLATFORM */
 }
```
FozzTexx, line 47: Shouldn't this method have moved to the new CP/M base class?

FozzTexx (commented):
It's a good start. The new base CP/M class farming out the device specific stuff for reading&writing is an improvement. The base class dealing with the differences between ESP32 and LWM on its own are great.

What I see though is that a lot of CP/M functionality hasn't entirely been moved up to the new base class. There's still a lot of CP/M stuff being handled in the bus subclasses where it should be already handled in the base class.

The `ep_*()` methods also don't seem quite right. To me they look like RunCPM hooks which the bus subclasses shouldn't need to care about. There's already `host_available()`, `host_read()`, and `host_write()` which the subclasses use to get data in&out of the CP/M base class in their bus specific methods.

PR #1405, opened 2026-06-30
Fix: check malloc() result in fnHttpClient to prevent null pointer crash

yazdan-iot:
## What this fixes

`fnHttpClient`'s constructor calls `malloc()` to allocate the response buffer but never checks if the allocation succeeded. If `malloc()` fails (e.g. low memory on the ESP32), `_buffer` stays `nullptr`.

Later, in `_httpevent_handler` (HTTP_EVENT_ON_DATA), the code does:
```cpp
memcpy(client->_buffer, evt->data, client->_buffer_len);
```
without checking for null, which would crash on the first data received over HTTP.

## Changes

- Added a null check after `malloc()` in the constructor, with a debug log if allocation fails.
- Added a guard in the `HTTP_EVENT_ON_DATA` handler to safely drop incoming data (instead of crashing) if `_buffer` is null.

## Testing

Built and tested using FujiNet-PC on Windows (MSYS2/CLANG64). Build succeeds with no errors, all existing tests pass.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# FujiNet afdsssssssssss
+# FujiNet
 
 A multi-function peripheral built on ESP32 hardware being developed for the multiple 8-bit systems
 
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# FujiNet
+# FujiNet afdsssssssssss
 
 A multi-function peripheral built on ESP32 hardware being developed for the multiple 8-bit systems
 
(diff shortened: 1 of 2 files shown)


tschak909:
What happened with that first commit?

yazdan-iot:
> What happened with that first commit?

Sorry, that was a mistake — I accidentally left some test text in README.md while getting familiar with the workflow, then committed it along with the actual fix. I've removed it in a follow-up commit. The README change isn't intentional, only the fnHttpClient.cpp fix is meant to be part of this PR.

FozzTexx (approved):

PR #1410, opened 2026-07-03
[adam] network: HTTP channel mode command + error code visibility

tschak909:
Two fixes to the ADAM network device, found while rounding out the Adam SmartBASIC extension (NHTTPMODE and ?Network Error trapping):

- `process_http()` now reads the channel mode from the packet and ACKs. The mode came from stale `cmdFrame.aux2` (only set by open) and the packet was never ACKed, hanging the ADAM in the resend loop.
- Error codes are now visible to the host: bus status polls and `'E'` are passive (`fromInterrupt`), so they no longer fire the lazy HTTP transaction — the bus poll was doing that inside its 300µs reply deadline. `get_error()`/`status()` report the preserved reason after a failed open, `json_parse` records `COULD_NOT_PARSE_JSON`, TCP accept clears a stale error on success.

Built and linked for `FUJINET_TARGET=ADAM`.


FozzTexx:
This looks an awful lot like the same change I just made to apple/iwm. Maybe all of those changes should be done to Adam too? Someone really needs to 1108 the network stuff.

https://github.com/FujiNetWIFI/fujinet-firmware/pull/1369/changes

tschak909:
@FozzTexx let's go ahead and get this in, and we'll refactor network soon.

FozzTexx:
image

DDP for testing:

[fnettest.ddp.zip](https://github.com/user-attachments/files/30065299/fnettest.ddp.zip)

FozzTexx (approved):

PR #1414, opened 2026-07-05
DriveWire/CoCo high score information support.

RichStephens:
High-score sectors for CoCo DSK images

Allows a DSK image to be mounted read-only while still permitting writes to a designated set of "high-score" sectors — so game code is protected but scores can still be saved.

How it works:

- A marker block at LSN 323 (track 17, sector 18 — unused by CoCo Disk BASIC) identifies up to 4 ranges of writable sectors. The format is the 4-byte magic FNHS, a version byte (0x01), a count, then pairs of (start LSN, count) as big-endian 16-bit values.
- On mount, MediaTypeDSK reads and parses this marker, storing the ranges. The _media_read_only flag is now set from the access mode before mount() is called.
- On a write to a read-only disk, the media layer checks whether the target block falls in a high-score range. If not, the write is rejected. If it is, the firmware temporarily reopens the file read-write via the fujiHost, performs the write and flush, then closes that handle and restores the original read-only one.
- fujiDevice passes the fujiHost pointer down to the disk device via the new set_media_host() call after a successful mount (CoCo only), so the media layer has what it needs to do the reopen.

### DIFF
--- a/lib/device/drivewire/disk.cpp
+++ b/lib/device/drivewire/disk.cpp
@@ -59,6 +59,7 @@ mediatype_t drivewireDisk::mount(fnFile *f, const char *filename, uint32_t disks
     if (_media)
     {
         strcpy(_media->_disk_filename,filename);
+        _media->_media_read_only = !(access_mode & DISK_ACCESS_MODE_WRITE);
         mt = _media->mount(f, disksize);
         if (mt == MEDIATYPE_UNKNOWN)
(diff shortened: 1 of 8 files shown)


FozzTexx (changes requested):
Don't put ifdefs in fujiDevice.

FozzTexx (approved):

PR #1423, opened 2026-07-07
Created SystemBusBase to define the contract between devices and system buses.

FozzTexx:
This introduces a common interface for accepting transactions ownership, receiving data, providing responses, and reporting completion or errors. This interface will replace device-owned transaction handling. Bus implementations are responsible for protocol-specific details while devices interact only through the transaction API.

This commit only defines the contract; bus implementations and device migration will follow.


tschak909 (approved):
Yup.

PR #1429, opened 2026-07-08
Convert Drivewire bus to SystemBusBase transaction interface

FozzTexx:
Move transaction state and handling from the device layer into Drivewire systemBus class, making the bus responsible for implementing the SystemBusBase contract.


tschak909 (approved):
Awesome.

PR #1430, opened 2026-07-08
[coco] Decouple virtualDevices from the transport layer

FozzTexx:
Introduce `FujiDWPacket`, a lazy-deserializing packet implementation that presents the same logical interface as `FujiBusPacket` while accommodating DriveWire's context-dependent packet format.

Packet fields are materialized on demand and cached, allowing parameters to be accessed in arbitrary order while preserving protocol ordering on the wire. Multi-byte parameters are transparently converted from DriveWire's big-endian encoding to native byte order.

This change also removes the remaining public transport bypass interfaces (`read()`, `write()`, `available()`, and `flush()`), eliminating the last direct dependencies between device implementations and the bus hardware.

Devices no longer perform bus I/O or implement DriveWire protocol mechanics. The bus layer is now solely responsible for transport, framing, and protocol handling, while devices operate exclusively on decoded packet abstractions.

### DIFF
--- a/fujinet_pc.cmake
+++ b/fujinet_pc.cmake
@@ -389,6 +389,7 @@ if(FUJINET_TARGET STREQUAL "COCO")
     list(APPEND SOURCES
 
     lib/bus/drivewire/drivewire.h lib/bus/drivewire/drivewire.cpp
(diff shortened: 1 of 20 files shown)


tschak909 (approved):
Ok.

PR #1431, opened 2026-07-09
[network-protocol][fs] Raw filename directory type.

tschak909:
Add 0x83 for raw enumerated directory type for things like wildcard copying. NOS uses this.
-Thom


RichStephens (approved):

RichStephens:
I dig it. :)

PR #1433, opened 2026-07-09
Add seek to fs http

tschak909:
This adds the ability for seek/tell (NOTE/POINT) to the HTTP protocol object, allowing random access.

Seeks now rebuild the client connection and set the range.


FozzTexx (approved):

idolpx:
Rebuilding the client slows things down significantly.  Look at the way I did it in MeatHttpClient.  I've had seeking working for years.  It falls back to closing and reopening the connection when seeking backwards if the server doesn't support range headers too.

PR #1434, opened 2026-07-09
[fujiHost][http] add random access with keep-alive to fujiHost HTTP/S.

tschak909:
* Adds a new fnFileHTTP class that can properly manage esp_http and mgHttp* to handle ranges, with keep-alive.
* Wire in fnFileHTTP to fujiHost
* Turn on keepalive for N: device so that notes-points don't have to rebuild connection.


tschak909:
@FozzTexx you ok with this?

FozzTexx (approved):

PR #1437, opened 2026-07-10
coco/drivewire: Ensure that failed reads get a retry

RichStephens:
This fixes the "coco boot goes to splash screen, then hangs" problem.


tschak909 (approved):
Oh cool. I did not see these in the spec.

FozzTexx (approved):

PR #1439, opened 2026-07-10
[iwm] Use `transaction_*` methods more often in iwm devices

FozzTexx:
More prep work to get direct bus handling out of the devices.


tschak909 (approved):
Yay

PR #1444, opened 2026-07-12
[network-protocol][http] defer to base class for WebDAV dirs.

tschak909:
The HTTP channel did not hold the size of the built directory entries parsed by the WebDAV object.

Fix was to defer to the base class, which does.


FozzTexx (dismissed):

tschak909:
SORRY!

RichStephens (approved):

PR #1447, opened 2026-07-12
transaction_get() fixes.

tschak909:
Drivewire - Set SSID
DriveWire - new_disk()
Drivewire - drivewireNetwork::set_prefix()
RS232 - rs232Disk::rs232_write()


RichStephens (approved):
Tested on Rev0000. Works!

PR #1449, opened 2026-07-12
Fix Atari NetStream TCP RX backpressure

mozzwald:
Limit TCP socket reads to available NetStream RX ring space so firmware stops draining and dropping bytes during bursts. Leave excess data in TCP where the window can provide backpressure. Also add DEBUG_NETSTREAM RX drop/high-water stats for hardware diagnosis.


tschak909 (approved):
Looks good.

PR #1451, opened 2026-07-12
[iwm] Introduce FujiIWMPacket

FozzTexx:
Introduce `FujiIWMPacket`, a packet implementation that presents the same logical interface as `FujiBusPacket` while accommodating the SmartPort context-dependent packet format.

Parameters are pulled from the SmartPort payload, allowing them to be accessed easily without needing to keep track of where the non-parameter bytes start. Multi-byte parameters are transparently converted from the Apple II little-endian encoding to native byte order.


tschak909 (approved):
Cool.

PR #1453, opened 2026-07-13
Created mixins for adding features to fujiDevice

FozzTexx:
Added a new ability to add features to fujiDevice via mixins which automatically handle command processing and don't need to be added to every single fujiDevice subclass.

Base64 and hash have already been converted to mixins. QR and appkeys will be converted soon.

Individual buses need to have been upgraded to use bus packet classes that conform to FujiBusPacket.


tschak909 (approved):
Ok, good.

FozzTexx:
Before:

image

After:

image

PR #1464, opened 2026-07-15
Add websocket protocols

tschak909:
* Implement a pair of protocols (WS and WSS) for connecting to web sockets
* Both implementations for esp32 and mongoose HTTP
* Tested against IRATA.ONLINE's websockify and echo.webtest.org


tschak909:
I've tested the crap out of this. @FozzTexx it's gtg.

FozzTexx (approved):

PR #1465, opened 2026-07-16
[iwm] Packet already removes the 3 leading bytes from the filename

FozzTexx:


tschak909:
Tested. It works. Approving. Thanks @FozzTexx

tschak909 (approved):

PR #1467, opened 2026-07-17
[adam] Introduce FujiAdamPacket

FozzTexx:
Introduce `FujiAdamPacket`, a packet implementation that presents the same logical interface as `FujiBusPacket` while accommodating the AdamNet context-dependent packet format.

Parameters are pulled from the AdamNet packet payload, allowing them to be accessed easily without needing to keep track of where the non-parameter bytes start


tschak909 (approved):
cool.

PR #1470, opened 2026-07-17
[network-protocol][Protocol] refactor translation.

tschak909:
Replace hard-coded compile time constants with saner code.


lib/network-protocol/Protocol.cpp
```diff
@@ -42,22 +42,64 @@ using namespace std;
...
+// Platforms whose native EOL is a CR/LF pair (serial / CP/M-style machines).
+#if defined(BUILD_RS232) || defined(BUILD_S100) || defined(BUILD_RC2014) || defined(BUILD_H89)
+#define NATIVE_EOL_CRLF
+#endif
+
+// Native EOL as a single character (single-EOL platforms only). Apple and the
+// CR/LF platforms handle their own EOL in translate_receive_buffer() and do not
+// use this value.
+#if defined(BUILD_ATARI)
+#define NATIVE_EOL_CHAR ATASCII_EOL
```
FozzTexx: Are we trying to future proof this? If so, `#ifdef`s really don't belong here. There should be a variable in the Protocol that the network class for each bus can assign itself instead of putting platform/bus dependencies in here.

lib/network-protocol/Protocol.cpp
```diff
@@ -163,51 +205,60 @@ fujiError_t NetworkProtocol::status(NetworkStatus *status)
...
     replace(receiveBuffer->begin(), receiveBuffer->end(), ASCII_BACKSPACE, ATASCII_DEL);
     replace(receiveBuffer->begin(), receiveBuffer->end(), ASCII_TAB, ATASCII_TAB);
-    #endif
+#endif
 
+#if defined(BUILD_APPLE)
+    // Apple II native EOL is CR. Modes 1 and 2 leave CR untouched; only mode 3
+    // acts, collapsing an incoming CR/LF pair down to a single CR.
+    if (translation_mode == NETPROTO_TRANS_CRLF)
+        util_replaceAll(*receiveBuffer, STR_ASCII_CRLF, STR_ASCII_CR);
```
FozzTexx: Why is CRLF translation Apple II only? Why wouldn't any other platform want to convert CRLF to the native EOL?

lib/network-protocol/Protocol.cpp
```diff
@@ -163,51 +205,60 @@ fujiError_t NetworkProtocol::status(NetworkStatus *status)
...
+
+#elif defined(NATIVE_EOL_CRLF)
+    // Native EOL is CR/LF; fold the incoming line ending into a CR/LF pair.
+    util_replaceAll(*receiveBuffer, network_line_ending(translation_mode), STR_ASCII_CRLF);
+
```
FozzTexx: Why would native CRLF preclude wanting to use the switch statement? Is it because the switch statement isn't necessary?

FozzTexx (commented):
This seems more complicated than necessary to me. Why isn't it a simple "native → requested network EOL mode" / "requested network EOL mode → native"?

lib/network-protocol/Protocol.cpp
tschak909: it shouldn't be, anymore.

lib/device/h89/network.cpp
```diff
@@ -612,6 +612,9 @@ bool H89Network::instantiate_protocol()
         return false;
     }
 
+    // Serial/CP-M machines use a CR/LF pair as their native EOL.
+    protocol->native_eol = "\x0d\x0a";
```
FozzTexx: Wouldn't it be better to use the `STR_`* constants defined by Protocol.h instead of hardcoding the characters?

RichStephens:
I'm a little worried about this breaking json queries like once happened before when messing with this stuff. Every value wiry returning a new line..

I semi I haven't had a chance to look closely at this code tho.

FozzTexx:
> I'm a little worried about this breaking json queries like once happened before when messing with this stuff. Every value wiry returning a new line..
> 
> I semi I haven't had a chance to look closely at this code tho.

I plan to run fnettest with the emulators. I should be able to remove this `#ifdef`/`#warning` from `tests/network.c` with these changes. Maybe `LCLTEST` would be a good test too since it uses JSON extensively.
```
#ifdef BUILD_ATARI
                     OPEN_TRANS_LF
#else
#warning "OPEN_TRANS_LF should work on ALL platforms"
                     OPEN_TRANS_NONE
#endif // BUILD_ATARI                                                                           
```

lib/device/h89/network.cpp
RichStephens: ALWAYS.

tschak909:
Ok, the fixed strings are replaced by human readable constants. I like turtles!

FozzTexx:
Almost works. Doesn't seem to handle CRLF to LF conversion. Test uses:
```
  err = network_open(NET_DEVICESPEC, OPEN_MODE_HTTP_GET, OPEN_TRANS_LF);
```

image

Works on Adam, Apple II, Atari, and CoCo though.

Also glad that the native EOL will be a variable. I think at some point down the road it makes sense to add a command to set the native line ending because when using OPEN_TRANS_LF the C compiler doesn't do what you expect.

image
image
image
image

FozzTexx (commented):
New NETCMD_SET_EOL command works fine.

CRLF to LF network translation still doesn't work.

Image

Image

FozzTexx (approved):
Image

PR #1477, opened 2026-07-20
Add lobby support for dragon

wdathing:
Add lobby support for dragon.
And update config program, add splash for dragon.


tschak909 (approved):
Thank you

PR #1479, opened 2026-07-20
Fix fujiversal prototype startup hang

RichStephens:
This PR originally DID pin the USB host tasks to core 1 on the Fujiversal DriveWire build

Symptom: On the CoCo Fujiversal prototype (RP2350 bridge + ESP32-S3 over USB), a cold boot intermittently hangs while loading the config program. The FujiNet serves the small CFGLOAD.BIN loader fine and the splash appears, but the following load of the larger CONFIG.BIN (≈77 sectors) stalls partway and the machine hangs at the splash. Resetting and re-running the boot works, and pre-powering the board over USB before switching the CoCo on avoids it entirely.

It no longer does that.  What it does instead is temporarily boost the priority of those tasks over the wifi task until such time as the wifi task obtains an IP address. Once that happens, the tasks' priorities are returned to normal.

This tests well with both startup and regular operation.

### DIFF
--- a/lib/hardware/ACMChannel.cpp
+++ b/lib/hardware/ACMChannel.cpp
@@ -140,9 +140,19 @@ void ACMChannel::begin()
     host_config.intr_flags = ESP_INTR_FLAG_LEVEL1;
     ESP_ERROR_CHECK(usb_host_install(&host_config));
 
+#ifdef PINMAP_FUJIVERSAL_DRIVEWIRE
+    // Keep USB servicing off core 0. On core 0 the WiFi stack preempts these
+    // tasks during cold boot and stalls the DriveWire byte stream mid-sector,
+    // hanging the CoCo's CONFIG load. Core 1 (only the serve loop lives there)
+    // stays clear of the WiFi bring-up storm.
+    BaseType_t task_created = xTaskCreatePinnedToCore(usb_lib_task, "usb_lib", 4096,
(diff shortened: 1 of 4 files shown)


lib/hardware/ACMChannel.cpp
```diff
@@ -140,9 +140,19 @@ void ACMChannel::begin()
...
 
+#ifdef PINMAP_FUJIVERSAL_DRIVEWIRE
+    // Keep USB servicing off core 0. On core 0 the WiFi stack preempts these
+    // tasks during cold boot and stalls the DriveWire byte stream mid-sector,
+    // hanging the CoCo's CONFIG load. Core 1 (only the serve loop lives there)
+    // stays clear of the WiFi bring-up storm.
+    BaseType_t task_created = xTaskCreatePinnedToCore(usb_lib_task, "usb_lib", 4096,
+                                          xTaskGetCurrentTaskHandle(),
+                                          USB_HOST_PRIORITY, NULL, 1);
+#else
```
FozzTexx: Don't put platform #ifdefs in here. This is an abstraction that should not be filled with board specific stuff.

FozzTexx (changes requested):
This broke running Lobby for me. Hangs trying to load the servers.

tschak909:
image

FozzTexx (approved):

PR #1480, opened 2026-07-22
[config] update coco and apple2 with joystick support.

tschak909:
* Add joystick nav support for APPLE2 and COCO.
Each tested and working on Go.


RichStephens:
LOL I just flashed and tested this in my bus prototype.

RichStephens (approved):

PR #1487, opened 2026-07-25
QR codes: fix Atari, add mixin for other platforms

nwah:
Fixes:
- No longer crashes when outputting on Atari
- Format of the to_binary is fixed to send full thing again

Adds:
- QRMixin to add support other platforms


lib/bus/rs232/FujiBusPacket.h
```diff
@@ -97,7 +97,9 @@ class FujiBusPacket
     fujiCommandID_t command() const { return _command; }
 
     std::uint32_t param(unsigned int index) const {
-        return _params[index].value;
+        // Commands may read more params than a client sent (e.g. QR encode's
+        // optional 3rd "shorten" param); treat absent params as zero.
+        return index < _params.size() ? _params[index].value : 0;
```
FozzTexx: This shouldn't blindly accept wrong values. It's supposed to crash on purpose. The location that is trying to request a parameter that doesn't exist needs to be checking if there's enough parameters.

lib/device/iwm/iwmFuji.cpp
```diff
@@ -392,10 +392,8 @@ void iwmFuji::iwm_ctrl_qrcode_encode(const iwm_decoded_cmd_t &cmd)
     Debug_printf("FUJI: QRCODE ENCODE\n");
     Debug_printf("QR Version: %d, ECC: %d, Shorten: %s\n", version, ecc_mode, shorten ? "Y" : "N");
 
-    std::string url = _qrManager.data;
-
     if (shorten) {
-        url = fnHTTPD.shorten_url(url);
+        _qrManager.data = fnHTTPD.shorten_url(_qrManager.data);
```
FozzTexx: Rip out all the QR stuff from iwmFuji. This should be handled by the QRMixin.

FozzTexx (changes requested):
I'm just a day or so away from making Atari use the mixins too. I just need to get the packet class added, which I have in another branch at the moment.

FozzTexx (dismissed):
It'd be nice to have Atari ready for mixins so that all the QR stuff could be removed from sio.cpp too, but it should be easy enough to change it later.

Thanks for beating me to the punch with this! It's been on my backlog for quite some time.

tschak909 (approved):
yay

PR #1492, opened 2026-07-28
Move PacketParam and PacketParamProxy into their own .h files

FozzTexx:
Move PacketParam and PacketParamProxy into their own .h files since they are shared by so many different bus packets.


tschak909 (approved):
Ok

PR #1493, opened 2026-07-28
atari: modem: add configurable answer delay to fix some BBS servers

mozzwald:


FozzTexx (dismissed):
Isn't this supposed to be some register that you set with AT commands that you have to go look up in the modem manual and page past 3 times before you find it? Web interface is too easy.

mozzwald:
I accidentally pushed another commit unrelated to modem which has now been added as a different PR. oops

tschak909 (approved):

PR #1494, opened 2026-07-28
atari: netstream: use ephemeral port for localhost udp, fix regression

mozzwald:
This is really only useful for fujinet-pc, unless you are running a game server on the esp32 also :grin:


mozzwald:
Added fix for Windows MSYS2 build. Also added a regression fix from FujioSIOPacket introduction that broke Netstream port.

tschak909 (approved):
Yay

PR #1504, opened 2026-08-03
Convert appkeys into a mixin

FozzTexx:
### opened as draft

* Removed appkeys from fujiDevice and moved them over to AppKeyMixin.


FozzTexx:
This will temporarily break appkeys on Lynx until the bus can be upgraded to support fujiDevice mixins, but I believe currently none of the apps are yet using appkeys? @shawnjefferson22 @mozzwald what do you think? Should this wait?

PR #1505, opened 2026-08-05
Make a single network protocol factory method.

FozzTexx:


tschak909 (approved):
makes sense.

idolpx:
cool... that's what NetworkProtocolFactory was for.  :)

PR #1508, opened 2026-08-08
Implement intellivision

tschak909:
Ok, here's the pico firmware, the test rom, and a couple of small bugfix changes on the RS232 BoIP side.


build-platforms/platformio-fujinet-rs232-s3.ini
```diff
@@ -4,8 +4,8 @@ build_bus      = rs232
 build_board    = fujinet-rs232-s3
 
 [env]
-upload_port = /dev/ttyACM0 ; Linux esp32s3
-monitor_port = /dev/ttyACM0 ; Linux esp32s3
+upload_port = /dev/ttyACM1 ; Linux esp32s3
```
mozzwald: most systems will likely use ACM0 for flashing. I don't think this should be changed

tschak909: oops, that should not have made it in, i'll revert that.

build-platforms/platformio-fujiversal-rs232.ini
```diff
@@ -3,6 +3,8 @@ build_bus = RS232
 build_platform = BUILD_RS232
 
 [env:fujiversal-rs232]
+upload_port = /dev/ttyACM0
+monitor_port = /dev/ttyACM0
```
FozzTexx, line 7: This is S3 stuff, doesn't belong here.

lib/device/adamnet/disk.cpp
```diff
@@ -45,7 +45,7 @@ void adamDisk::reset()
 }
 
 mediatype_t adamDisk::mount(fnFile *f, const char *filename, uint32_t disksize,
-                            disk_access_flags_t access_mode, mediatype_t disk_type)
+                            disk_access_flags_t access_mode, mediatype_t disk_type, fujiHost *host)
```
FozzTexx, line 48: I don't know why you've added an argument that nothing is using. Maybe this should be its own PR?

lib/hardware/ACMChannel.cpp
```diff
@@ -125,6 +125,48 @@ static void newDevForwarder(usb_device_handle_t usb_dev)
...
+
+    // Test Line Coding commands: Get current line coding, change
+    // it 9600 7N1 and read again
+
+    cdc_acm_line_coding_t line_coding;
+    ESP_ERROR_CHECK(cdc_acm_host_line_coding_get(cdc_dev, &line_coding));
+
+    line_coding.dwDTERate = 9600;
+    line_coding.bDataBits = 7;
+    line_coding.bParityType = 1;
```
FozzTexx, line 147: ACMChannel fixes should be its own PR

FozzTexx (commented):
I don't know why there's H89 stuff in here.

The pico firmware seems to have a lot of strange stuff in it, like FAT filesystem support. Disk and file stuff should be handled on the FujiNet side, not on the pico. Plus tons of comments that say "don't redistribute this source."

tschak909:
Split into reviewable PRs:

- #1510 -- `[rs232][hardware]` ACMChannel: recover from USB CDC re-attach
- #1511 -- `[rs232][network]` clear devicespec buffer on open
- #1512 -- `[fujiDevice]` fail MOUNT_IMAGE when mount() returns MEDIATYPE_UNKNOWN
- #1513 -- `[rs232]` ROM media type: push cartridge images to the RP2040 (stacked on #1512)
- #1514 -- `[pico][intellivision]` vendored RP2040 cartridge firmware -- **DO NOT MERGE**, blocked on license grants from Andrea Ottaviani (PiRTO II) and Robin Edwards (A8PicoCart-derived storage layer); see `pico/intellivision/PROVENANCE.md` in that PR

Also fixed along the way: the mediatype signature change here had widened `DISK_DEVICE::mount()` on all eleven platforms to carry a `fujiHost*` that only rs232 uses. #1512/#1513 replace that with a `fujiDevice::mount_media()` virtual hook instead, so the other ten platforms (sio/iec/iwm/adamnet/comlynx/cx16_i2c/rc2014/s100spi/h89/drivewire) are untouched. Dropped the stray `client/h89/*.COM` binaries and the local `/dev/ttyACM0` sdkconfig hardcode, neither of which belongs upstream.

Closing this in favor of the above.

PR #1509, opened 2026-08-08
Implement Clipboard feature

nwah:
Initial implementation of a bi-directional clipboard feature. Allows easy sending text or binary data between the host computer and another computer. Still would like to implement custom device handler for BASICs that makes it easy to list/load programs through this.

Screenshot 2026-08-08 at 5 48 00 PM


nwah:
Works well enough for now

tschak909 (dismissed):

tschak909 (approved):

PR #1510, opened 2026-08-09
[rs232][hardware] ACMChannel: recover from USB CDC re-attach

tschak909:
## Summary
- `begin()`'s wait loop stopped consuming `device_connected_sem` the moment it succeeded once, so a disconnect/replug (e.g. the RP2040 resetting) was never noticed and required a full ESP32 reboot to recover
- Extracts `openDevice()` out of `begin()`, hoists `dev_config` to a member so it can be reused, and adds a long-lived `ACM-reconnect` task that keeps listening for `device_connected_sem` after the first connection
- Enables USB hub support in the `fujiversal-rs232` sdkconfig (`CONFIG_USB_HOST_HUBS_SUPPORTED` / `HUB_MULTI_LEVEL` / downstream port recovery delay), needed for the same USB topology this targets

Split out of #1508 (see that PR for context).

## Test plan
- [ ] `pio run -e fujiversal-rs232` builds
- [ ] Unplug/replug the RP2040 mid-session on real hardware and confirm the link recovers without an ESP32 reboot


FozzTexx:
I'm pretty sure this fixes a problem both @nwah and @RichStephens have experienced so they should test this one.

RichStephens:
Doesn't ring a bell. What problem might that be?

FozzTexx:
> Doesn't ring a bell. What problem might that be?

Pushing reset on the pico itself or unplugging the USB cable into the pico meant the ESP32 had to be rebooted too.

RichStephens:
Huh. Never noticed. Well, good fix.

RichStephens (approved):

PR #1511, opened 2026-08-09
[rs232][network] clear devicespec buffer on open

tschak909:
## Summary
- Stale bytes from a previous, longer devicespec could leak into the next `deviceSpec` string if the new one was shorter
- One-line `memset` fix in `create_devicespec()`

Split out of #1508 (see that PR for context).

## Test plan
- [ ] `pio run -e fujinet-rs232-s3` (or fujiversal-rs232) builds
- [ ] Open a network device with a long devicespec, then a shorter one, and confirm no stale trailing bytes appear


FozzTexx (approved):
The packet already knows how long the payload is, but meh, not worth bothering with refactoring to pass the payload length since I'm about to 1108 the network devices this week anyway.

Approved.

PR #1515, opened 2026-08-10
add device password + web UI to manage it

nwah:
This is a prerequisite for potentially sensitive features, like an appkey viewer/manager.

It stores the password salted + hashed in flash, so should be relatively safe. However the current plan is to just use the stubbed out function to protect certain web ui routes with Basic Auth, so not super secure but better than nothing.

Screenshot 2026-08-09 at 10 30 07 PM


nwah:
Don't mind the "SEC🗻URITY"

tschak909 (approved):

PR #1522, opened 2026-08-11
Replace intv with minty

tschak909:
Replace the pico firmware with Minty, and some small fixes to ROM handling so that the DBC handoff works to it.


FozzTexx:
This seems to have other PRs mixed into it?

tschak909:
I will try to peel these apart into minty and rom handling PRs, and close this one.

tschak909:
There's so much stuff happening in parallel that it gets tangled. I'm implementing both sides at the same time, and there's constant tweaking on both sides to make things work.

tschak909:
Splitting this into two focused PRs so they can review independently:
- #1524 — vendorize Minty, replace old `pico/intellivision/` firmware
- #1525 — ESP32-side ROM media-type handling (`MEDIATYPE_ROM`)

(One commit from this branch, `0c1615050`, is dropped in the split — it added a size/extension heuristic that `a0991d7bd` (now #1525) fully replaces, so keeping it would just be net-zero churn.)

Closing this in favor of the two above.

PR #1525, opened 2026-08-11
[rs232] ROM media type: push cartridge images to the RP2040

tschak909:
## Summary
Adds `MEDIATYPE_ROM`, detected by extension (`.rom`/`.bin`/`.int`/`.itv`) in `MediaType::discover_mediatype()`. These formats mean "load straight into cart memory and reset" on every current `BUILD_RS232` platform (Intellivision `.bin`+`.cfg` / Intellicart `.rom`, Atari 2600 `.bin`, CoCo cartridge images), so detecting by extension alone is unambiguous. Replaces `rs232Disk::mount()`'s `disksize==8192||16384||32768` heuristic, which misfired on any file of those exact sizes regardless of extension and missed every other ROM size (e.g. `battleship.bin` at 17730 bytes).

`MediaTypeROM` (`lib/media/rs232/diskTypeROM.{h,cpp}`) streams the file to the RP2040 over FujiBus, addressed to `FUJI_DEVICEID_DBC`: a same-named `.cfg` sibling first (best-effort, so the RP2040 knows the memory mapping), then the ROM itself, each as `NETCMD_OPEN`/`WRITE`.../`CLOSE` on its own DBC stream id. `NETCMD_CLOSE` on the ROM stream triggers the actual boot.

`MediaTypeROM` needs a `fujiHost*` to open that `.cfg` sibling through the same host the ROM came from (TNFS, SD, ...), but `fujicore_mount_disk_image_success()` is shared by every platform and `lib/device/fujiDevice/` bans `#ifdef BUILD_*` (`tests/check_no_build_ifdefs.py`) — platform customization goes through a subclass override instead. Adds a protected virtual `fujiDevice::mount_media()` hook whose base implementation is the original 5-arg `mount()` call (used unmodified by sio/iec/iwm/adamnet/comlynx/cx16_i2c/rc2014/s100spi/h89/drivewire); `rs232Fuji` overrides it to pass `host` through to `rs232Disk::mount()`, the only platform that gained a trailing `fujiHost*` parameter.

## Scope
Split out of #1522, which mixed this in with an unrelated `pico/` vendor import (Minty firmware, now in #1524). This PR touches only `lib/` + `fujinet_pc.cmake`.

## Verification
- `tests/check_no_build_ifdefs.py lib/device/fujiDevice` passes clean
- Confirmed byte-identical to the ROM-media portion of #1522 (diffed against the exact set of files this commit touches)
- Build note: `fujinet-rs232-rev0` and `fujinet-rs232-s3` currently fail to build on `origin/master` itself (unrelated `PIN_ADAMNET_RESET` / `-Werror=format` issues in adamnet code) — verified pre-existing by building `origin/master` directly with none of these changes applied

### DIFF
--- a/fujinet_pc.cmake
+++ b/fujinet_pc.cmake
@@ -479,6 +479,7 @@ if(FUJINET_TARGET STREQUAL "RS232")
 
     lib/media/rs232/diskType.h lib/media/rs232/diskType.cpp
     lib/media/rs232/diskTypeImg.h lib/media/rs232/diskTypeImg.cpp
+    lib/media/rs232/diskTypeROM.h lib/media/rs232/diskTypeROM.cpp
 
     lib/device/rs232/apetime.cpp lib/device/rs232/apetime.h
     lib/device/rs232/disk.cpp lib/device/rs232/disk.h
--- a/lib/device/fujiDevice/fujiDevice.cpp
+++ b/lib/device/fujiDevice/fujiDevice.cpp
@@ -517,7 +517,7 @@ success_is_true fujiDevice::fujicore_mount_disk_image_success(uint8_t deviceSlot
     // We need the file size for loading XEX files and for CASSETTE, so get that too
     disk.disk_size = host.file_size(disk.fileh);
     DISK_DEVICE *disk_dev = get_disk_dev(deviceSlot);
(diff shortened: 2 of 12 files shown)


nwah (approved):
Will need to expand/revamp a bit for ROM types, but headed in the same direction I was

PR #1526, opened 2026-08-11
Add `halfDuplex` option to ESP32UARTChannel

FozzTexx:
`halfDuplex` setting automatically suppresses echo on single-wire serial connections.

Will convert Adam and Lynx in their own PRs.


tschak909 (approved):
That is awesome.

PR #1527, opened 2026-08-11
[adam] Put UART into half duplex mode

FozzTexx:
Put UART into half duplex mode and let it suppress echo instead of discarding all the echo manually.


tschak909 (approved):

benjamink:
Appears to work for me.  I was able to boot FCS from `apps.irata.online`

IMG_6823

PR #1529, opened 2026-08-12
[intv] add back in the fujiversal-intv.

tschak909:
The reconstruction continues. Add back in the intv board definition, as well as the guard that will ultimately allow upstream detection of the RP2xxx from the ESP32S3 and thus auto-flashing (this is coming in another PR).


FozzTexx (dismissed):

tschak909:
Apologies. working with @gtortone showed that octal mode psram needed to be added back in (it was part of the original PR but got missed).

FozzTexx (approved):

tschak909:
thank you good sir.

PR #1531, opened 2026-08-12
[build] add merged build for intv.

tschak909:
This adds the ability for INTV to produce a build that has the RP2040 binary embedded within it, to be ultimately used by the auto-flashing feature.


tschak909:
the odd CI-CD blongle aside, it's ok to review. :)

FozzTexx:
My thoughts:

How would we approach this if we also considered making the solution reusable across other platforms?

Right now, this seems very focused on accomplishing this one specific thing as quickly as possible. I think it's worth considering what happens when we want to add support for MSX or CoCo or KC85 or NES or whatever comes next. How much of this would have to be unwound or duplicated to make it work with another board?

I'm not necessarily suggesting that we need to solve all of that right now, but I think it's worth asking whether we're building something that the next person can reuse, or whether we're just solving this particular problem and leaving the next person to figure out how to make it work for their board.

I think that's an important part of the goal here, even if it isn't necessary to get this particular use case working.

tschak909:
I completely agree. This code will definitely change.

Let's put this in for now.

We can try to bring it over and implement on our other two platforms, MSX, and COCO, this should be a good enough sampling to see what it would take to extend this. I only guarded this one to minimize potential damage while active development is going on all over the place.

PR #1533, opened 2026-08-13
[adam] Bus handles MN_STATUS instead of devices directly writing to bus

FozzTexx:


tschak909 (approved):
that makes sense. cool.

PR #1534, opened 2026-08-13
[pico][intellivision] fix mailbox size.

tschak909:
INTV was failing because the DBC packets were being truncated.

### DIFF
--- a/pico/intellivision/firmware/include/fuji_mailbox.h
+++ b/pico/intellivision/firmware/include/fuji_mailbox.h
@@ -100,6 +100,20 @@
                                                    // 256-byte host-slot block and the
                                                    // 240-byte AdapterConfigExtended
 
+// fujibus_transact()'s raw USB receive buffer (fuji_mailbox_service()'s
+// rxbuf) has to be sized separately from FUJI_MB_RX_MAX above: that constant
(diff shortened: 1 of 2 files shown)


FozzTexx:
Why not just use FujiBusPacket.cpp straight from the FujiNet firmware? That class is used verbatim on the Fujiversal firmware and openMSX and was designed to be entirely standalone and not be tied to FujiNet firmware.

tschak909:
@FozzTexx because the code is plain C and not C++.

FozzTexx:
> @FozzTexx because the code is plain C and not C++.

It's still gcc that can compile C++ though? Linker doesn't care if it's C++ or C.

tschak909:
I'm trying it. It's at 96% RAM usage, so this may overflow as a result.

tschak909:
Yup, confirmed. Using the FujiBusPacket.cpp directly costs an additional 4K of RAM, which is already tight (baseline is at 96%).

sorry.

-Thom

FozzTexx (approved):

FozzTexx:
> Yup, confirmed. Using the FujiBusPacket.cpp directly costs an additional 4K of RAM, which is already tight (baseline is at 96%).

Lousy C++.

PR #1537, opened 2026-08-14
[lynx] Introduce FujiLynxPacket, comlynx inherits from SystemBusBase

FozzTexx:
* comlynx systemBus inherits from SystemBusBase
* FujiLynxPacket is serialized/deserialized by systemBus
* Devices no longer read/write packets directly
* `comlynx_process()` is passed packet
* lynxFuji forwards command processing to fujiDevice for standard commands


mozzwald (approved):
Works now on both real hardware and emulation 👍

PR #1541, opened 2026-08-16
[iec] Introduce FujiIECPacket

FozzTexx:
* iecFuji forwards standard command handling to fujiDevice
* Removed all the BASIC support from iecFuji since nothing used it and it needs to be reworked
* All flibtests pass


FozzTexx:
IMG_3540

tschak909:
Awesome :)

tschak909 (approved):
awesome.

PR #1549, opened 2026-08-18
[rs232] Make the N: filesystem commands work

RichStephens:
mkdir, rmdir, delete, rename, lock and unlock always failed on this bus. No request ever reached the server and nothing was logged to say why.

process_fs() now reads the devicespec off the wire and instantiates a protocol for it, instead of reusing whatever an earlier open had left behind. Each command carries the path it acts on, so it has to be read before the operation can be pointed at the right target.

### DIFF
--- a/lib/device/rs232/network.cpp
+++ b/lib/device/rs232/network.cpp
@@ -1185,11 +1185,31 @@ void rs232Network::rs232_set_timer_rate(int newRate)
 
 void rs232Network::process_fs(const FujiBusPacket &packet)
 {
+    /* The FS commands are sent without aux bytes, so there may be no
+     * parameter to read. param() throws when the index is absent. */
+    fileAccessMode_t access = static_cast<fileAccessMode_t>(
+        packet.paramCount() > 0 ? packet.param(0) : 0);
+
+    SYSTEM_BUS.transaction_accept(TRANS_STATE::WILL_GET);
+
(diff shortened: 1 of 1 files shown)


lib/device/rs232/network.cpp
```diff
@@ -1185,11 +1185,31 @@ void rs232Network::rs232_set_timer_rate(int newRate)
...
+
+    parse_and_instantiate_protocol(access);
+
     // Make sure this is really a FS protocol instance
     NetworkProtocolFS *fs = dynamic_cast<NetworkProtocolFS *>(protocol.get());
     if (!fs)
     {
-        SYSTEM_BUS.transaction_error();
+        // transaction_error() was already called from
+        // parse_and_instantiate_protocol()
```
FozzTexx: This comment makes no sense. If error was already called then why call it again?

FozzTexx (commented):
Let me see what this does to `ncopy` which has been using the FS stuff.

RichStephens:
> Let me see what this does to `ncopy` which has been using the FS stuff.

Weird.  Because I couldn't get any network_fs functions to do anything at all without it.

lib/device/rs232/network.cpp
RichStephens: Revised it to be clearer.

FozzTexx (approved):
I can make the firmware crash using `fnshare` without this fix, so seems like it helps.

PR #1550, opened 2026-08-18
Fix fujinet_pc.cmake so that git commit ID matches build

FozzTexx:
This fixes it so that the GET_EXTENDED command reports a git commit ID that matches what was actually built against. The firmware version printed by FUJINET.SYS matches what I see when I do `git rev-parse --short HEAD`.

Try this out @RichStephens and see if it works for you too.

### DIFF
--- a/fujinet_pc.cmake
+++ b/fujinet_pc.cmake
@@ -606,16 +606,14 @@ endif()
 
 # Version file
 # run build_version_pc.py to generate ${CMAKE_BINARY_DIR}/include/build_version.h
-add_custom_command(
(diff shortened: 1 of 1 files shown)


RichStephens:
Will try it when I get back home. Had to go get my auto emissions test done.

RichStephens:
Looks good to me: 
```
richstephens@fujibook:~/fujinet-firmware$ git rev-parse --short HEAD
c90b2a61
```

image

RichStephens (commented):
Looks good, and works!

RichStephens (approved):

PR #1551, opened 2026-08-18
Turn on checks for Adam and Lynx LWM

FozzTexx:


tschak909 (approved):
sure, might as well. :)

PR #1552, opened 2026-08-18
[http] Accept a PUT body on the POST data channel

RichStephens:
A channel opened as PUT that delivers its body through the SET_POST_DATA channel mode had that body discarded, leaving nothing to send, and the request was then abandoned rather than issued. Both fujinet-lib and fujinet-lib-experimental route network_http_put through network_http_post, which uses that channel mode, so every PUT made through either library was affected.

Widen the method check in write_file_handle_send_post_data to accept PUT as well as POST. Both write paths append to the same postData buffer and differ only in which method they admit, so this closes the gap that plain PUT fell through, without changing behavior for POST or PUT_H.

### DIFF
--- a/lib/network-protocol/HTTP.cpp
+++ b/lib/network-protocol/HTTP.cpp
@@ -652,7 +652,7 @@ fujiError_t NetworkProtocolHTTP::write_file_handle_set_header(uint8_t *buf, unsi
 
 fujiError_t NetworkProtocolHTTP::write_file_handle_send_post_data(uint8_t *buf, unsigned short len)
 {
-    if (httpMethod != HTTP_METHOD::POST)
+    if (httpMethod != HTTP_METHOD::POST && httpMethod != HTTP_METHOD::PUT)
     {
         error = NDEV_STATUS::INVALID_COMMAND;
         return FUJI_ERROR::UNSPECIFIED;


FozzTexx:
I recall that we had to undo this recently. I know there was something broken with PUT testing that I just fixed too, let me make sure that's been pushed.

RichStephens:
> I recall that we had to undo this recently. I know there was something broken with PUT testing that I just fixed too, let me make sure that's been pushed.

Actually, I noticed that, too.  PUT was actually hitting the POST endpoint and failing, and because of some other bugs where failures still get reported as success, it wasn't failing the test.  (I have a branch for that fix, too, waiting on this).

FozzTexx:
The PUT tests weren't using the put endpoint correctly. Try this fix against master, then see if it passes against this PR.

https://github.com/FozzTexx/fujinet-lib-experimental/pull/64

RichStephens:
> The PUT tests weren't using the put endpoint correctly. Try this fix against master, then see if it passes against this PR.
> 
> [FozzTexx/fujinet-lib-experimental#64](https://github.com/FozzTexx/fujinet-lib-experimental/pull/64)

All FNETTEST tests pass with your PR on my firmware (with this PR), on coco,  atari, apple2, and msdos. Didn't test Adam yet. I'll get back to adam before I check in my latest tests.

FozzTexx:
> All FNETTEST tests pass with your PR on my firmware (with this PR), on coco, atari, apple2, and msdos. Didn't test Adam yet. I'll get back to adam before I check in my latest tests.

But do they pass before your change? Because if the tests pass on master then either this PR 1552 isn't necessary, or the tests are broken. I'd lean towards the tests being broken...

RichStephens:
> > All FNETTEST tests pass with your PR on my firmware (with this PR), on coco, atari, apple2, and msdos. Didn't test Adam yet. I'll get back to adam before I check in my latest tests.
> 
> But do they pass before your change? Because if the tests pass on master then either this PR 1552 isn't necessary, or the tests are broken. I'd lean towards the tests being broken...

They APPEAR to pass before my change, because there's that "silent fail" issue that I was talking about.  The POST endpoint rejects the PUT open, which fails, but the open doesn't report back failure.  So, if I put in another PR, the one that actually reports all the "silent failures", then the test (which definitely needed to be fixed) would have failed.

FozzTexx:
> The POST endpoint rejects the PUT open, which fails, but the open doesn't report back failure.

I wonder if there's anything the test can do to report FAIL on current master.

RichStephens:
> > The POST endpoint rejects the PUT open, which fails, but the open doesn't report back failure.
> 
> I wonder if there's anything the test can do to report FAIL on current master.


OK, it's more complicated than I thought. You changed the test from a PUT test, to a PUT_H test.  A PUT test would have internally failed, silently, because it would have gone through the true "PUT" path and and failed to actually PUT anything, but because of the "silent failure" issues, and the fact that the test was not checking the response, it would have appeared to pass. If the test had checked the response, it probably would have failed, because httpbin sends back what you sent when you do a POST.

Now, because you're actually testing "PUT_H" - put with headers - that takes the same path as POST, it wouldn't fail at all with current firmware. 

I think maybe the test needs to have both a PUT_H test and a true "PUT" test, as well as my fix here (and probably changes to ensure that the response is checked?)

Some of the filesystem functions use a true PUT, which is what was failing for me in my filesystem tests.

FozzTexx (approved):
Fixes the new PUT test!

PR #1556, opened 2026-08-19
[intv] Support paged, oversized and low-mapped homebrew cartridges

tschak909:
Homebrew INTV cartridges failed to load in three unrelated ways:

- `BOOT_MAX_SEGS 16` rejected any `.cfg` with more mapping lines (frantic 4 has 17, Cloudfire 24, Pacmanthology 30). Now 48.
- `PAGE` mapping lines were silently mis-parsed — the 3-value `sscanf()` matches them and drops the qualifier, so every page painted over the last. `mm_add()` and cartridge.c's `$xFFF = $A5n` decode already handled paging; only the network parser didn't.
- `.rom` segments below `$4000` were rejected, since the decoder wrote at the bus address and would clobber the CONFIG image the console is executing mid-transfer. `$2000-$2FFF` is ordinary cart space (Maria, Space Intruders, unlucky_pony_2024 all start there). Both formats now stage sequentially above CONFIG and map by delta.

Parser split out of `fujinet.c` into `bootmap.c` (no pico-sdk/tusb/Cartridge), so `host_test/test_bootmap.c` can run real cartridge files through the same calls the USB handler makes and dump the resulting map — the sample set is checkable without flashing. Parser now matches `load_cfg()`: case-folded, comment-stripping, PAGE-aware, and streaming rather than buffering the whole `.cfg`.

`NETCMD_OPEN` now carries the stream size, so an oversized ROM is refused before it's pulled over TNFS and the progress bar is exact; older peers send the id alone and still work. New `TOOBIG` and `BANKSW` reason codes replace failures that previously reported something misleading.

A `.cfg` pairing `jlp` with ROM across `$8000-$9FFF` was a hard `ERR_MAILBOX` reject; JLP's window shadows that ROM anyway, as under jzintv, so those segments are dropped and the cart boots (Pacmanthology needs this). `[memattr]` RAM under the window is still a real collision and still rejected.

## Testing

Host test covers all ten sample cartridges plus synthetic cases for uppercase sections, CRLF, no trailing newline, and the size-guess/`NOMAP` paths:

```
cd pico/intellivision/firmware/host_test
gcc -Wall -Wextra -I../include -o /tmp/test_bootmap test_bootmap.c ../src/bootmap.c ../src/memory.c ../src/utils.c
/tmp/test_bootmap "$HOME/tnfs/INTV/sample roms/"*.bin "$HOME/tnfs/INTV/sample roms/"*.rom
```

`pirto_ii_default` (97.85% RAM, down from 99%), `fujicard` and `fujiversal-intv` all build. **Not yet tested on hardware.**

Cloudfire and Pacmanthology remain too large for `pirto_ii_default`'s `cart.ROM[]` — an RP2040 SRAM ceiling, not a bug; they now fail as `TOOBIG` rather than `REJECTED`.

https://claude.ai/code/session_01J3nfDqKnhQZGyHgayymWSB

### DIFF
--- a/lib/media/rs232/diskTypeROM.cpp
+++ b/lib/media/rs232/diskTypeROM.cpp
@@ -64,15 +64,13 @@ static bool push_stream(fnFile *f, uint16_t stream_id, uint32_t expected_size)
 {
     uint8_t buf[DISK_SECTORBUF_SIZE];
 
-    char open_payload[5] = {
-        (char)(uint8_t)stream_id,
-        (char)(uint8_t)(expected_size & 0xFF),
-        (char)(uint8_t)((expected_size >> 8) & 0xFF),
-        (char)(uint8_t)((expected_size >> 16) & 0xFF),
-        (char)(uint8_t)((expected_size >> 24) & 0xFF),
-    };
+    struct { uint8_t id; u32le_t size; } open_hdr;
+    static_assert(sizeof(open_hdr) == 5, "OPEN header must not be padded");
+    open_hdr.id = (uint8_t)stream_id;
+    open_hdr.size = expected_size;
+
     auto reply = SYSTEM_BUS.sendCommand(FUJI_DEVICEID_DBC, NETCMD_OPEN,
-                                        std::string(open_payload, sizeof(open_payload)));
+                                        std::string((const char *)&open_hdr, sizeof(open_hdr)));
     if (!reply || reply->command() != FUJICMD_ACK)
     {
         Debug_printv("MediaTypeROM: failed to open DBC stream %u (%lu bytes)\n",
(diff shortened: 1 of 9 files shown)


lib/media/rs232/diskTypeROM.cpp
```diff
@@ -43,26 +43,40 @@ void MediaTypeROM::status(uint8_t statusbuff[4])
...
+//
 // Always sends NETCMD_CLOSE so the RP2040's stream state doesn't wedge; a
 // failed transfer's CLOSE carries a 0x01 abort payload so partial data
 // isn't booted.
 static bool push_stream(fnFile *f, uint16_t stream_id, uint32_t expected_size)
 {
     uint8_t buf[DISK_SECTORBUF_SIZE];
 
-    uint8_t open_payload = (uint8_t)stream_id;
+    char open_payload[5] = {
```
FozzTexx:
```
struct {
  uint8_t id;
  u32le_t size;
};
```

tschak909: implemented.

tschak909:
@FozzTexx suggested fix applied.

FozzTexx (approved):

PR #1557, opened 2026-08-19
Add XML channel mode, and general text sanitization

nwah:
Gumbo turns out to be too HTML-focused to serve as general-purpose SGML parsing/querying. This adds a separate XML channel mode built on the tinyxml2 library already vendored in the firmware, plus helpers for folding HTML entities (`&amp;`) and UTF-8 characters (`”`, `—`) down to plain ASCII, usable when querying JSON, HTML **or** XML.

Rebuilt on top of the 1108'd N: device (#1624). The original version of this PR predated that work and carried ~700 lines of near-identical wiring across five buses; `NDevice`'s `NParser` abstraction absorbs all of it, so **no per-bus device file is touched** other than the one Atari override below.

Rebased onto master now that the FNSGML→FNHTML rename has landed as #1642, split out of this PR at review request.

### Review items addressed

- **`query_flags()` hook removed from `NDevice`.** The Atari-only behavior lives in Atari: `sioNetwork` overrides `fujidev_set_query()` directly, the way it already overrides `fujidev_set_prefix()`.
- **The leftover `query_param` variable is gone** from `NDevice::fujidev_set_query`, along with any comment about how other buses decode packets. The base handler just calls `fujicore_set_query(query, 0)`.
- **Comment dropped from the `_parser` null check** — it is a plain bug fix, not a note about a subclass.
- **`H89Network.h` / `rc2014Network.h` untouched**, being obsolete.
- **`network_data.h`** conflict resolved by rebasing: #1641 deleted the file.

### What's here

| | |
|---|---|
| `lib/fnxml` | `FNXML` over tinyxml2, plus a standalone XPath-subset query engine |
| `lib/fntext` | `fn_sanitize` (entity decoding + UTF-8→ASCII transliteration) and `fn_query_flags.h` |
| `lib/device/NDevice/XMLParser` | `PARSER::XML = 3` on `NET_SET_PARSER` |

The XPath subset covers steps, `//`, `*`, `[n]`, `[@attr]`, `[@attr="v"]`, `[text()="v"]`, and a trailing `@attr` to return an attribute instead of element text. Namespace prefixes are part of the name, but a bare local name also matches a prefixed element, so `creator` finds `dc:creator`. Malformed queries match nothing rather than throwing — they arrive verbatim from an 8-bit host — and the descendant search is an explicit queue, not recursion, so a deeply nested hostile document can't blow the stack.

Output mode lives in bits 4-5 of the query param, the same bits in all three modes. `VERBATIM` is 0, so a client that sends no output mode behaves exactly as it does today; `ASCII` (0x10) decodes entities and folds to printable ASCII.

### Two master bugs this uncovered

Reviewers should look at these two commits closely; they are fixes to master, not part of the feature, and `OUTPUT_ASCII` was unreachable without them:

- **`[network] stop NET_QUERY clobbering the parser's query flags.`** `setQuery()` passed a hardcoded `0` into `setReadQuery()`, which assigns `_queryParam` and then resolves — so flags set via `NET_SET_PARAMETER` were wiped before `processString()` ran. Every query flag was dead on the NDevice buses, `JSON_REMAP_CHARS` and `JSON_DELETE_SGML_TAGS` included. Separately, the `NET_SET_PARAMETER` validation rejected anything above 2, so `JSON_DELETE_SGML_TAGS` (0x04) could never be set either.
- **`[atari] restore the NET_QUERY query-flags byte on SIO.`** Pre-1108, sioNetwork passed DAUX2 of `NET_QUERY` straight into `setReadQuery()`. Restored by overriding `fujidev_set_query()` in `sioNetwork`. The shared handler in `NDevice` passes no flags: only `FujiSIOPacket` decodes params from the command frame, so only SIO can read one at that point without eating the head of the query string.

### Testing

`fn_sanitize_tests` (108 assertions), `fnxml_query_tests` (11 subcases) and a new `nquery_output_mode_tests` that drives `XMLParser` against a canned feed through the real sequence — set flags, query, read — and asserts ASCII folds while verbatim doesn't. That pair is the tripwire for the hardcoded-0 bug: reinstating it fails 17 assertions. It also covers match iteration to exhaustion, the namespace-prefix fallback, and a malformed document reporting a parse error.

All six fujinet-pc targets build and pass (ATARI 8/8, others 6/6). ESP32 `fujinet-atari-v1` builds. Cost on ESP, measured against master: **+23.5 KB flash (3270867 → 3294915), +64 bytes static RAM** — tinyxml2 was vendored but dead-stripped, and is now live.

### Known gap, not addressed here

`fujidev_set_query` no longer strips trailing EOL bytes or a leading `N:`/`N#:` prefix from the query, both of which the pre-1108 SIO handlers did. That affects JSON and HTML equally and wants its own issue; for now clients must send the query bare and NUL-terminated.


tschak909 (dismissed):

FozzTexx:
This couldn't wait until *after* the network is 1108? 🤪

tschak909:
nothing ever happens when we want it to.

nwah:
> This couldn't wait until _after_ the network is 1108? 🤪

You just gotta take a long time to merge it 😉

tschak909:
What are we doing with this one guys?

FozzTexx:
I think it's stalled waiting for the N: Device 1108.

FozzTexx:
N: Device 1108 is in. #1624

tschak909:
am attempting to have an intern put this back in shape now that N: is 1108'd. Stay tuned.

lib/device/h89/H89Network.h
```diff
@@ -11,7 +11,6 @@
 
 #include "Protocol.h"
 #include "fnjson.h"
-#include "fnsgml.h"
```
FozzTexx, line 14: H89Network is obsolete, don't change it.

lib/device/rc2014/rc2014Network.h
```diff
@@ -11,7 +11,6 @@
 
 #include "Protocol.h"
 #include "fnjson.h"
-#include "fnsgml.h"
```
FozzTexx, line 14: rc2014Network is obsolete, don't change it.

lib/device/NDevice/NDevice.cpp
```diff
@@ -427,6 +429,21 @@ void NDevice::fujidev_set_prefix(const FUJI_COMMAND_PACKET &packet)
 error_is_true NDevice::fujicore_set_query(const std::string &query, uint8_t parseFlags)
 {
     error_is_true err = error_is_true(false);
+
+    // rs232Network reaches this helper through its own handler, which has no
+    // null check of its own.
```
FozzTexx: Not checking if `_parser` is null is a bug, not a problem of the subclass. Comment not needed.

lib/device/NDevice/NDevice.h
```diff
@@ -144,6 +145,17 @@ class NDevice : public virtualDevice
...
+    /**
+     * Query flags accompanying NET_QUERY, or 0 for "none sent".
+     *
+     * Defaults to 0 on purpose. Only SIO may override it: its params come from
+     * the already-decoded command frame, whereas AdamNet/ComLynx/IEC/IWM decode
+     * params by erasing bytes off the front of the packet payload and DriveWire
+     * reads them straight off the wire -- so reading a param here would eat the
+     * first bytes of the query string.
+     */
+    virtual uint8_t query_flags(const FUJI_COMMAND_PACKET &packet) const { return 0; }
```
FozzTexx: This is backwards. Only Atari does it different so Atari should be overriding `fujidev_set_query()` to do it differently. NDevice should not be adding Atari-only compatibility code.

lib/network-protocol/network_data.h
```diff
@@ -11,19 +11,19 @@
...
     JSON     = 1,
-    SGML     = 2,
+    HTML     = 2,
 } channelMode_t;
 
 struct NetworkData {
     std::unique_ptr<NetworkProtocol> protocol;
     std::unique_ptr<FNJSON> json;
-    std::unique_ptr<FNSGML> sgml;
+    std::unique_ptr<FNHTML> html;
```
FozzTexx: This file was supposed to be deleted, see PR #1641

FozzTexx (commented):
If SGML is becoming HTML then the rename should be a PR all on its own.

lib/device/h89/H89Network.h
tschak909: Reverted — untouched here now. The only edit was dropping an unused `#include "fnsgml.h"`, which came in with the rename; the rename is now #1642, and it leaves both files alone as well.

lib/device/rc2014/rc2014Network.h
tschak909: Same — reverted, untouched here and in #1642.

lib/device/NDevice/NDevice.cpp
tschak909: Comment dropped; the null check stays.

lib/device/NDevice/NDevice.h
tschak909: Agreed — `query_flags()` is gone. `sioNetwork` now overrides `fujidev_set_query()` directly, next to its existing `fujidev_set_prefix()` override, and reads DAUX2 there. `NDevice` is back to passing no flags; the only thing left in the base class is a comment saying why (the other buses erase params off the front of the payload, so a param read there eats the head of the query string) — that seemed worth keeping since a bare `0` is what made this look like an oversight in #1624.

lib/network-protocol/network_data.h
tschak909: Rebased onto master, so the edit to it is simply gone along with the file.

lib/device/NDevice/NDevice.cpp
```diff
@@ -435,7 +450,10 @@ error_is_true NDevice::fujicore_set_query(const std::string &query, uint8_t pars
 
 void NDevice::fujidev_set_query(const FUJI_COMMAND_PACKET &packet)
 {
-    uint8_t query_param = 0; //packet.param(1);
+    // No flags off the packet here: AdamNet/ComLynx/IEC/IWM decode params by
+    // erasing bytes off the front of the payload and DriveWire reads them off
+    // the wire, so a param read here would eat the head of the query string.
+    uint8_t query_param = 0;
```
FozzTexx: How packets work is outside the scope of a device class and the comment isn't relevant. Just delete the variable entirely, it was a leftover from the refactor work and didn't get cleaned up.

lib/device/NDevice/NParser.cpp
```diff
@@ -56,7 +56,8 @@ NetworkStatus NParser::status()
 
 error_is_true NParser::setQueryParam(uint8_t param)
 {
-    RETURN_ERROR_AS_TRUE();
+    _queryParam = param;
+    RETURN_SUCCESS_AS_FALSE();
```
FozzTexx: This is wrong. The base NParser has nothing to set a query for and should always return an error. `_queryParam` is something that belongs in a subclass.

lib/device/NDevice/NParser.h
```diff
@@ -11,6 +11,11 @@ class NParser
     // Last parse() result, so STATUS can say why the channel is empty. Set by each parse().
     nDevStatus_t _parseError = NDEV_STATUS::SUCCESS;
 
+    // Flags for the next query (see fntext/fn_query_flags.h). Held here rather
+    // than only in the wrapper because setQuery() has to hand them back to
+    // setReadQuery() on every call, which would otherwise reset them to 0.
+    uint8_t _queryParam = 0;
+
```
FozzTexx: Belongs in a subclass that uses it.

lib/device/NDevice/NDevice.cpp
tschak909: Gone — the variable and the comment both. The base handler just calls `fujicore_set_query(query, 0)` now.

I trimmed the comment on the `sioNetwork` override to match: it says only that DAUX2 carries the flags on SIO, with the cross-bus reasoning dropped for the same reason.

Also rebased onto master now that #1642 has landed, so the `SGML` names are back to `HTML` here.

lib/device/NDevice/NParser.h
tschak909: This is the pull that doesn't end...it just goes on and on my friend...

FozzTexx: Yah those comments got stuck somehow and I thought they had gone in with one of the reviews this morning. 🫤

tschak909: Gone from `NParser` entirely, rather than moved down — `FNJSON`/`FNHTML`/`FNXML` already store this byte (`setQueryParam()` writes it), so the copy was redundant on top of being in the wrong class. Each wrapper gets a `queryParam()` getter and `setQuery()` reads it back:

```
-    _json.setReadQuery(query, 0);
+    _json.setReadQuery(query, _json.queryParam());
```

`NParser.h` and `NParser.cpp` are now untouched by this PR, and the subclass `setQueryParam()` overrides are back to what they were on master. Same test coverage, still passing.

If you would rather each subclass just hold its own copy, say the word and I will do that instead.

FozzTexx (approved):

lib/device/NDevice/NParser.cpp
tschak909: Agreed, and already the case as of the last push — this thread was the one I missed, but the fix covers it. `NParser::setQueryParam()` is back to `RETURN_ERROR_AS_TRUE()`, exactly as on master, and `NParser.cpp`/`NParser.h` are no longer touched by this PR at all. The flags live only in `FNJSON`/`FNHTML`/`FNXML`, which already owned them.

PR #1564, opened 2026-08-20
[adam] Remove unused query_device/adamQueryDevice

FozzTexx:


tschak909 (approved):
Sorry. It won't be missed.

PR #1567, opened 2026-08-21
Handle 0-length app keys (used to clear an app key)

RichStephens:


FozzTexx:
There's another simpler way to do that, let me look in rs232Fuji.cpp.

FozzTexx:
```
packet.dataAsString().value_or("")
```

FozzTexx:
Maybe can do `packet.data().value_or({})` ? Not sure, C++ is so inconsistent.

RichStephens:
> ```
> packet.dataAsString().value_or("")
> ```

dataAsString() on the adamnet, comlynx, and iwm buses crashes on a zero-length write.

It reads the data field without checking whether anything is in it:
```
auto d = data();
return std::string(reinterpret_cast<const char *>(d->data()), d->size());
```
When the field is empty, the pointer and length it reads are garbage.

That's why dataAsString().value_or("") wouldn't have fixed it.  It would crash dataAsString() before value_or can hand back the empty string.

FozzTexx:
> dataAsString() on the adamnet, comlynx, and iwm buses crashes on a zero-length write.

That's a whole lotta bugs in the packet code. ☹️

But does C++ reject the `packet.data().value_or({})` idea? I didn't really want to use a string, I just knew there was some magic "or" thing in rs232Fuji so I pasted what was there.

RichStephens:
Here's a simpler fix. What do you think?

FozzTexx (approved):

FozzTexx:
> Here's a simpler fix. What do you think?

Wow, I can't believe C++ was willing to cooperate! Usually trying to use those convenience methods results in half a page of casting and recasting and splitting objects up to recombine them.

PR #1569, opened 2026-08-23
fujiDevice handles the processing of commands

FozzTexx:
* The switch statement in fujiDevice subclasses now only handles bus specific commands
* Created a map in fujiDevice with all the common functions and their methods
* map allows single source of truth for checking if command is supported


FozzTexx:
IMG_3662
IMG_3661
IMG_3652
IMG_3649
IMG_3646
IMG_3656

tschak909 (approved):
good.

PR #1574, opened 2026-08-26
[intv][pico][firmware] Update embedded CONFIG ROM: colour + selection-bar flicker fix

tschak909:
Re-embeds the Intellivision CONFIG ROM from `fujinet-config/intv`, covering two upstream changes:

1. **Colour pass** (`fujinet-config` 7b3b2ae) — the CONFIG screens got a real palette: FGBG backgrounds on the host and info screens, a colour-stack selection bar and file-type glyphs in the file browser.
2. **Selection-bar flicker fix** (`fujinet-config` 49ce329) — every screen with a selection bar repainted its entire list on each cursor move. `#BACKTAB` is the live STIC display list with no shadow buffer, and an NTSC frame only affords ~50 IntyBASIC cell writes, so those repaints (~320 cells on the host list, a *blocking mailbox round trip per row* on the WiFi list) were scanned out half-finished for several frames running. The file browser additionally spent a >1-frame window carrying 2 of its 4 colour-stack advance bits, shifting everything below the bar one stack position.

## Firmware-side changes

`config.cfg`'s `[mapping]` moved with the image size, so all four `mm_add()` tuples in `fuji_config_map()` were re-transcribed — cross-checked by regexing both files and comparing tuples, not by eye, since a stale entry maps a segment short and the code past it is silently unreachable:

| before | after |
|---|---|
| `$1000-$1F9C -> $6000` | `$1000-$1ECA -> $6000` |
| `$1F9D-$2B49 -> $D000` | `$1ECB-$2BC2 -> $D000` |
| `$2B4A-$3036 -> $F000` | `$2BC3-$30AF -> $F000` |

`FUJI_STAGE_BASE` is unchanged — the image is `0x30B0` words and `0x3100` remains the next 256-word boundary past it, which the existing `_Static_assert` confirms at compile time.

## Verification

- `pirto_ii_default` configures, compiles and links clean.
- The new 24928-byte `config.bin` appears verbatim in `Minty_pirto_ii_default.elf`, and the previous 24686-byte image is gone.
- Upstream, the ROM was exercised under jzIntv with a probe driving both list screens: the host list cycles its selection bar with every glyph intact and correctly uppercase-folded, and the file browser walks its bar with all four colour-stack advance bits in phase in every frame.

Not yet exercised on real PiRTO II hardware.

https://claude.ai/code/session_01Ew7HzDQ3PHtp1uTh7BaT1D


FozzTexx (dismissed):

tschak909:
sorry. one more change. This fixes the flickering while moving the selection bar in various screens by drastically optimizing the output path!

FozzTexx (approved):

PR #1582, opened 2026-08-28
[rs232] Add MediaTypeDSK for raw 8"/5.25" floppy disk images

deltecent:
## What

Adds `MediaTypeDSK`, a new RS232 media type that serves bit-exact raw
sector dumps of real 8" and 5.25" floppies to the host, addressed by
head/track/sector with the disk format named explicitly per access. The
`.DSK` image stays a pure sector dump — no header, no sidecar, never
modified — and geometry is always explicit (firmware-baked or
host-declared), never inferred from the image bytes.

## Why

The existing RS232 disk path (`MediaTypeImg`) is flat 512-byte LBA. Real
vintage floppies aren't shaped that way: they're addressed by
head/track/sector, use non-512 sector sizes (e.g. 128 B), and can vary
geometry across the medium (the Tarbell DD boundary where track 0
differs from the rest). A host running a disk-controller emulator
(WD177x/FD1771-class) or an original disk BIOS needs to read and write
the raw dump exactly the way the floppy controller did — by CHS with the
correct per-track sector layout. The flat-LBA type can't represent any
of that, so today those images either can't be served faithfully or have
to be mangled into an LBA layout the host doesn't expect.

`MediaTypeDSK` closes that gap: it hands the host authentic raw floppy
dumps addressed the native way, and — via host-declared custom geometry —
covers arbitrary formats (including 5.25" and self-formatting media)
without baking every one into firmware or touching the image on disk.

## How

This follows the same media-type architecture the Atari SIO disk layer
already uses — a base `MediaType` with per-format subclasses — where
`MediaTypeDSK` is the RS232 analog of Atari's geometry-aware
`MediaTypeATX` (as opposed to the flat `MediaTypeATR`/`MediaTypeImg`).

- **New `MediaTypeDSK`** (`lib/media/rs232/diskTypeDSK.*`), a
  `MediaType` subclass beside `MediaTypeImg`/`MediaTypeROM`. Its
  `decode_sector()` resolves CHS + fmttype to a byte offset via a
  region-based format table. `fmttype` packs a 4-bit format index
  (bits 0–3), a whole-track access-mode bit (bit 6), and reserved bits
  that are refused if set.
- **Five baked 8" formats**: IBM SD/DD, Altair, FDC+, Tarbell DD.
- **Host-supplied custom geometry**: index `0x0F` (`FMT_CUSTOM`) resolves
  against a geometry the host declares once via `DISKCMD_SET_GEOMETRY`
  (`0x47`) and then addresses on every access — so a WD177x/FD1771-class
  controller emulator or a self-formatting BIOS can serve arbitrary
  geometry (odd sector sizes/counts, mixed density, double-sided,
  per-side differences), including 5.25", with no baked format. Regions
  accumulate via an append flag; side order is a declared property
  (`head_major`); the staging buffer heap-grows when a custom track
  exceeds the 6656 B baked bound.
- **Base `MediaType`** gains `decode_sector()` / `set_geometry()` /
  `sector_buffer()` / `sector_buffer_size()` virtuals, all
  default-preserving for the flat path, so the floppy type owns its
  addressing and a track-sized staging buffer without growing the
  512-byte buffer shared by Img/ROM.
- `.DSK` routes to `MEDIATYPE_DSK` in `discover_mediatype()`.

## Scope / safety

- **RS232 only.** No change to other platforms.
- **Behavior-preserving for `MediaTypeImg`/`MediaTypeROM`** by
  construction — the new base virtuals default to the existing flat-LBA
  behavior.

## Testing

- FujiNet-PC RS232 `dist` build green.
- New doctest suite (`tests/DiskTypeDSKTests.cpp`): 26 cases /
  607 assertions, covering the offset math for all five baked formats
  (incl. the Tarbell DD region boundary), the custom-geometry cases
  (uniform 512/1024 B, interleaved vs head-major double-sided, per-side
  track 0, custom-Tarbell == baked), plus a `make_dsk.py` image
  generator and a flat-IMG regression guard.
- ESP RS232-S3 firmware builds and validated on hardware.

Design doc: `docs/rs232-dsk-media.md`.

Design, format decisions, direction, on-hardware validation, and steering me
out of the occasional rabbit hole by Patrick Linstruth (@deltecent). Implemented
with [Claude Code](https://claude.com/claude-code).

### DIFF
--- a/lib/media/rs232/diskType.h
+++ b/lib/media/rs232/diskType.h
@@ -39,6 +39,7 @@ enum mediatype_t
     MEDIATYPE_UNKNOWN = 0,
     MEDIATYPE_IMG,
     MEDIATYPE_ROM,
+    MEDIATYPE_DSK,
     MEDIATYPE_COUNT
 };
 
@@ -93,6 +94,32 @@ public:
 
     // Always returns 128 for the first 3 sectors, otherwise _sectorSize
     virtual uint16_t sector_size(uint32_t sectornum);
+
+    // Resolve an addressed sector from bus-agnostic parameters (the device
+    // layer extracts these from its command packet). Default: params[0] is a
+    // linear sector (LBA), matching every existing flat image. MediaTypeDSK
+    // overrides this to decode head/track/sector/fmttype into a byte offset.
+    virtual uint32_t decode_sector(const uint32_t *params, unsigned count)
+    {
+        return count ? params[0] : 0;
+    }
+
+    // Declare a geometry from bus-agnostic parameters. Default: ignore -- Img/ROM
+    // have fixed geometry. MediaTypeDSK overrides this to build a runtime
+    // custom format (FMT_CUSTOM) from host-supplied regions.
+    virtual void set_geometry(const uint32_t *params, unsigned count)
+    {
+        (void)params;
+        (void)count;
+    }
+
+    // Staging buffer the device reads into / writes out of. Default: the existing
+    // 512-byte base member. A type whose transfers can exceed that (MediaTypeDSK,
+    // whose track-mode reads move a whole track -- up to 6656 bytes for an IBM 8"
+    // DD track) overrides these to supply its own larger buffer, so
+    // MediaTypeImg/MediaTypeROM never carry the cost.
+    virtual uint8_t *sector_buffer() { return _disk_sectorbuff; }
+    virtual uint32_t sector_buffer_size() { return DISK_SECTORBUF_SIZE; }
 
     virtual void status(uint8_t statusbuff[4]) = 0;
 
--- a/docs/rs232-dsk-media.md
+++ b/docs/rs232-dsk-media.md
@@ -1,0 +1,1361 @@
+# RS232 Firmware — `MediaTypeDSK` (CP/M 8" floppy disk images)
+
+This document is the implementation plan for a new RS232 media type,
+`MediaTypeDSK`, that serves **bit-exact raw dumps of real 8" (and 8"-class)
+floppy disks** — IBM, Altair, FDC+ — to a CP/M host, addressed the way the host's
+disk BIOS already thinks: by **head / track / sector**, with the disk's **format** named
+explicitly on every access.
+
+The design is pinned by these hard constraints:
+
+- **The `.DSK` image is a pure sector dump. No header, no sidecar, no modification —
+  ever.** These images are copies of physical media and must be writable back to a real
+  floppy unchanged. Nothing may be prepended, appended, or stored beside them.
+- **The host's CP/M BIOS is the geometry authority.** Its read/write sector routines know
+  the disk's format, so every command carries `head / track / sector / fmttype`.
+- **The ESP32 holds the format table.** `fmttype` indexes a firmware-baked table of known
+  floppy geometries; the FujiNet turns CHS into a byte offset into the raw dump. Geometry
+  is **never inferred** from the image (not from a header, not from size, not at mount).
+- **`MediaTypeImg` is not touched.** `MediaTypeDSK` is a new `MediaType` subclass beside
+  `MediaTypeImg` and `MediaTypeROM` — it inherits the base `MediaType`, not `MediaTypeImg`,
+  so none of the flat-512 behavior comes along. The one shared device edit (§6) is
+  behavior-preserving for `MediaTypeImg` by construction.
+
+The design body (§1–12) is the plan of record. Reference material this design draws on —
+the existing RS232 disk layer (Appendix A), prior-art geometry-aware media types elsewhere
+in the codebase (Appendix B), and the design alternatives these constraints rule out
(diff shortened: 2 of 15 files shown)


deltecent:
An example system using the new `MediaTypeFloppy` is available here: **https://github.com/deltecent/8mb-fujinet**

It's Altair 8Mb CP/M 2.2 booting entirely through a FujiNet RS232 — a whole-track FDC+ READ/WRITE (`fmttype=0x43`) `MediaTypeFloppy` transport driving a three-stage boot chain (ROMFUJI → BOOTFUJI → BIOSFUJI), with a reproducible offline build. Note it requires the firmware from this PR.

lib/media/rs232/diskType.h
```diff
@@ -4,6 +4,7 @@
 #include <stdint.h>
 #include "fnio.h"
 #include "global_types.h"
+#include "../../bus/rs232/FujiBusPacket.h"
```
FozzTexx: The disk/media image shouldn't know anything about bus packets. The image represents the actual contents of the media; bus packets are just one way a device can request operations on those contents. Those are separate layers. The bus code should take a packet, figure out what operation is being requested, and then ask the media image to perform that operation. Making the image aware of bus packets couples the media representation to a particular bus protocol and makes it impossible to cleanly use the same image through a different interface. The media image should deal with things like blocks, sectors, files, etc., not how those requests happened to arrive over the bus.

deltecent:
Thank you for the review, and for taking the time to point this out. The feedback is genuinely appreciated.

I've pushed a fix (`4fe58396f`): the `MediaType` addressing hooks no longer take a `FujiBusPacket`. The device layer (`rs232Disk`) now parses the packet and passes the media type only plain values — `decode_sector(const uint32_t *params, unsigned count)` and `set_geometry(const uint32_t *params, unsigned count)` — and `diskType.h` drops the `FujiBusPacket` include entirely, so the media layer no longer depends on the bus protocol. The unit tests call the media API directly with plain arrays now (and no longer link the bus code at all).

Wire format and behavior are unchanged; 26/607 assertions still pass. I've also merged the latest master in to clear the conflict from the `fujiCommandID` enum-class change.

deltecent:
Follow-up: I rebased this branch onto the latest master and dropped an unrelated disk-reopen commit that is being handled separately, so the force-push rewrote the commit SHAs. The branch is now just two commits — the floppy media type and the bus-agnostic fix — and it merges cleanly.

deltecent:
Force-pushed a rename: **`MediaTypeFloppy` → `MediaTypeDSK`** (and the whole `Floppy`/`FLOPPY` identifier family with it).

**Why:** "Floppy" doesn't actually distinguish this type — the IMD type in #1585 also serves floppy images. `DSK` names it after the `.DSK` extension it handles, and lines up with the existing `MediaTypeIMD` / `MediaTypeROM` / `MediaTypeImg` convention.

**What changed (identifiers only; descriptive "floppy" prose left as-is):**
- `MediaTypeFloppy` → `MediaTypeDSK`, `MEDIATYPE_FLOPPY` → `MEDIATYPE_DSK`
- `FloppyFormat`/`FloppyRegion`/`floppy_formats`/`floppy_fmt_t` → `DSKFormat`/`DSKRegion`/`dsk_formats`/`dsk_fmt_t`
- `FLOPPY_*` bounds macros, `_floppy_buff`, the include guard → `DSK_*` / `_dsk_buff`
- Debug log labels `"FLOPPY …"` → `"DSK …"`
- Files: `diskTypeDSK.{h,cpp}`, `DiskTypeDSKTests.cpp`, test target `diskdsk_tests`, `docs/rs232-dsk-media.md`

`MEDIATYPE_DSK` stays a plain append after `MEDIATYPE_ROM` (no pinned value) — it coexists fine with #1585's `MEDIATYPE_IMD`; they just take the next two slots in whatever order they land.

While rewriting for the rename I also **squashed the two commits into one**, which drops the intermediate "added with `FujiBusPacket`, then removed it" step — the final commit already carries the bus-agnostic media layer. No behavior or wire-format change; this is purely a naming pass.

Re-verified on the final commit: FujiNet-PC RS232 build clean, `diskdsk_tests` 26 cases / 607 assertions pass, and the `coding-standard.py` format check is clean.

deltecent:
On-hardware confirmation: flashed this commit to the RS232-S3 and booted CP/M off a `.DSK` — works fine, no behavior change from the rename.

The renamed log label is live, and the offset math checks out end to end:

```
DSK R fmt=3(FDC+) TRK 0/2/1 4384 @2240
```

- `DSK R` — the new label (was `FLOPPY R`)
- `fmt=3(FDC+)` + `TRK` — host sent `0x43` (`FMT_FDCPLUS | FMT_MODE_TRACK`), the whole-track FDC+ mode
- `4384` — a full FDC+ track = 32 × 137 B ✓
- `@2240` — 0x2240 = 4384 × 2 = track 2 in the flat dump ✓

deltecent:
Will post again later.

PR #1585, opened 2026-08-29
[media][imd] add an IMD mediatype for subclasses to use.

tschak909:
Adds an ImageDisk (`.imd`) media type and the RS232 adapter that serves it, so a CP/M-era floppy image mounts with its real geometry instead of being flattened into a uniform-sector `.img`. Mixed density (128-byte FM track 0, 512-byte MFM data tracks) presents as it did on the original drive.

- `lib/media/IMDImage.{h,cpp}` — the format core. Deliberately **not** a `MediaType`: every platform declares its own incompatible one, so each platform's adapter subclasses its own and forwards here. Takes an `fnFile *` and no platform header, because the ESP builds glob `lib/media/*.cpp` into every target. Written from the IMD.TXT spec (public domain); the ImageDisk program source was not consulted.
- `lib/media/rs232/diskTypeIMD.{h,cpp}` — the RS232 adapter, plus `MEDIATYPE_IMD` wiring.
- `lib/FileSystem/fnio.h` — honour a build-forced `FNIO_IS_STDIO` so the core unit-tests against `std::FILE`.

Two new ctest binaries: `imdimage_tests` (parser, sector and linear read/write, record types, deinterleave, truncation) and `mediatypeimd_tests` (the adapter).

### Testing

Rebased onto `master` at `209b0e0e1`; resolved a conflict in `rs232Disk::mount()` where the ROM case had changed.

- `make atari-lwm` — **13/13 ctest passed**, including both new suites.
- `./build.sh -b` for `fujinet-rs232-rev0` — SUCCESS.
- **Not flashed. No hardware verification.** No real drive, no Z80 machine, no round-trip against ImageDisk itself. The core is covered by the two suites; what is untested is whether a real CP/M client likes the geometry the adapter reports.


lib/media/imdFnFileSource.h
```diff
@@ -0,0 +1,65 @@
+#ifndef IMD_FNFILE_SOURCE_H
+#define IMD_FNFILE_SOURCE_H
+
+// fnFile-backed ImdSource, for platform adapters.
+//
+// Split out of imdImage.h so the core stays free of fnio: fnFile is FileHandler
+// on some targets and std::FILE on others, and dragging that into the core would
+// make every caller target-specific.
```
FozzTexx: I don't think this is true at all. It was my understanding `fnFile` works on all targets. There's shouldn't be a need to make a yet another file handle type, and especially not a file handle that is for a single media class.

```
#ifdef FNIO_IS_STDIO
  #include <cstdio>
  #include <unistd.h>  // for fsync
  typedef std::FILE fnFile;
#else
  #include "fnFile.h"
  typedef FileHandler fnFile;
#endif
```

lib/media/imdImage.h
```diff
@@ -0,0 +1,219 @@
...
+    virtual bool     read_at(uint32_t off, void *dst, uint32_t len) = 0;
+    virtual bool     write_at(uint32_t off, const void *src, uint32_t len) { (void)off; (void)src; (void)len; return false; }
+    virtual uint32_t size() = 0;
+    virtual bool     flush() { return true; }
+    virtual bool     writable() const { return false; }
+};
+
+class ImdCursor;
+
+class ImdImage
```
FozzTexx: "IMD" is not a word, it's an initialism. If the first letter is capitalized then all letters in "IMD" should be capitalized. Similarly, if the "i" is lowercase, then it should be "imd".

lib/media/imdFnFileSource.h
tschak909: You're right, and the comment was wrong. `fnio.h` typedefs `fnFile` unconditionally and the `fnio::` wrappers are uniform, so `fnFile *` already is the portable handle — there was no reason for a second abstraction, least of all one invented for a single media class.

Fixed in 08caedce7: `imdFnFileSource.h` and the `ImdSource` interface are gone. `IMDImage::open()` now takes `(fnFile *f, uint32_t disksize, bool writable)`, matching the `MediaType::mount(fnFile *, uint32_t)` signature every platform media type already uses, so an adapter just forwards its own `mount()` args. The bounds checks `ImdSource` used to own moved into `imd_read_at`/`imd_write_at` in the .cpp, shared by the cursor and the image, so they still live in one place.

One knock-on: the tests inherit `-D${FUJINET_BUILD_PLATFORM}` from `CMAKE_CXX_FLAGS`, which resolves `fnFile` to `FileHandler` and would drag the whole FileSystem stack into a unit test. `fnio.h` now has an `#ifndef FNIO_IS_STDIO` guard so a build can force stdio — which its own comment already contemplated — and the test target does that and writes its fixture to a temp file. The `FileHandler` path is still compile-checked by the ESP builds, which glob `lib/media/*.cpp` into every target.

lib/media/imdImage.h
tschak909: Agreed. Fixed in 08caedce7: `IMDImage`, `IMDStatus`, `IMDSectorInfo`, `IMDTrackInfo`, `IMDCursor`, `IMDStatus::NotIMD`, and the test's `IMD` fixture builder. That also lines up with `MediaTypeATX`/`MediaTypeWOZ`/`MediaTypeIMG` elsewhere in the tree. `tests/ImdImageTests.cpp` renamed to `tests/IMDImageTests.cpp`.

I left the all-lowercase `imd_status_str` / `imd_mode_*` free functions and the `imdImage.{h,cpp}` filenames as they are, since the first letter is lowercase there — same shape as `mediaTypeIMG.h` and `diskTypeAtx.h`. Say the word if you'd rather those changed too.

FozzTexx: The case of the filenames should match the case of the class name.

Can we get this hooked up to RS232 at a minimum so it can be tested?

tschak909:
Both done.

**Filename case** — `61add82dc` renames `imdImage.{h,cpp}` to `IMDImage.{h,cpp}` as its own commit so git records it as `R100`. Note this makes them the only uppercase-first filenames under `lib/media/` (every sibling is `diskTypeImg.h`, `mediaTypeIMG.h`, …) — happy to have it that way, just flagging that the directory is otherwise uniform.

I kept the *directory's* scheme for the new adapter: `lib/media/rs232/diskTypeIMD.{h,cpp}` holding `MediaTypeIMD`, matching `diskTypeImg.h`→`MediaTypeImg` and `diskTypeROM.h`→`MediaTypeROM`. Say the word if you'd rather that pair were `MediaTypeIMD.{h,cpp}` too.

**Hooked up to RS232** — `1c328a601` adds the adapter, `56ee2b6ca` adds a doctest suite for it.

`MediaTypeIMD` presents the image with its **native geometry**: drive sector N is IMD LBA N, and `sector_size()` returns the real per-sector size, so a 128-byte CP/M sector is served as 128 bytes. `rs232_read()` already sends `readcount` bytes and `rs232_write()` already sizes the receive from `sector_size()`, so variable and sub-512-byte sectors work with no change to the bus. `discover_mediatype()` picks it up from the `.imd` extension, so mounting one from the WebUI just works; `IMAGE_EXTENSION` and new-disk creation are untouched.

Decisions I'd like a second opinion on:

- **Sector numbering is 0-based.** `data/BUILD_RS232/autorun.img` is a raw FAT12 image whose boot sector is at byte 0, and `MediaTypeImg::_sector_to_offset(n)` is `n * 512` with no `-1`. It goes through a single `sector_to_lba()` helper, so it's a one-line change if that's wrong.
- **Writability is a constructor argument.** `rs232Disk::mount()` is the only place `access_mode` is in scope, and it had been accepting and discarding it. Widening the `mount()` virtual again — it was just widened in #1525 for `host`/`filename` — seemed like the wrong trade for a parameter one subclass wants.
- **Sectors larger than `DISK_SECTORBUF_SIZE` refuse the mount** rather than overrunning the 512-byte `_disk_sectorbuff` in `rs232_write()`. Raising that to 1024 would cover most of the remaining CP/M images at about +4 KB across all eight slots — cheap, but it changes the buffer for every media type, so I left it out of this PR.
- **Data-error and deleted-address-mark records still deliver their contents**, with the condition flagged in the controller status. An error reply carries no payload (`sendReplyPacket` only copies `if (ack && data)`), so returning an error would give the client nothing at all rather than the best-effort data — and that's a large fraction of real archived disks.
- **`format()` fails** instead of copying `MediaTypeImg`'s synthesized empty bad-sector map. IMD records are variable length and written in place; reporting a successful format of an archival image without touching it seemed worse than an error.
- `_percomBlock` gained an `= {}` initializer. It had none while every protected member does, and `DISKCMD_PERCOM_READ` sends all twelve bytes raw.

**Testing.** `tests/MediaTypeIMDTests.cpp` — 11 cases, 156 assertions. The adapter deliberately includes nothing from `lib/device` or `lib/bus`, so the media sources alone link into a doctest target. Verified: PC `RS232` and `ATARI` targets build and link; all four ctest suites pass; the ESP `fujiversal-intv` (ESP32-S3, `BUILD_RS232`) build succeeds with `MediaTypeIMD::mount(FileHandler*, …)` in the firmware, which covers the `fnFile`-as-`FileHandler` path the unit tests can't reach.

For hardware acceptance the useful image is an `.imd` made *from* an MSX 720K/360K disk — uniform 512-byte sectors, 9 spt — which should mount and boot exactly like the current `.img` path. An arbitrary CP/M IMD will mount and log its geometry correctly but won't be readable by a client that speaks FAT12.

**Four pre-existing things I found and deliberately did not touch**, since they change behaviour for the current client:

- `DISKCMD_STATUS` falls through to `rs232_write(packet.param(0), true)` alongside `DISKCMD_WRITE` in `lib/device/rs232/disk.cpp`, and nothing else calls `rs232Disk::rs232_status()` — so `MediaType::status()` is unreachable on this bus. The adapter implements it (it's pure virtual) and it's the first RS232 media type to honour the read-only TODO at `disk.cpp:101`, but none of that is observable today.
- `MediaTypeImg::read()`/`write()` bound with `sectornum > _disk_num_sectors`, one past the end.
- `rs232_format()` passes an uninitialized `responsesize` to `transaction_send()`.
- `rs232Disk::unmount()` closes the file but leaves `_disk` alive and non-null. `MediaTypeIMD` overrides `unmount()` to drop its borrowed handle first; `MediaTypeImg` would `fseek` a null handle on a stray read.

Happy to split any of those into their own PR if you want them fixed.

tschak909:
`IMDImage` is the ready-to-use part — it owns everything IMD: parse, geometry, per-sector read/write, cyl/head/sector-ID lookup, and a linear byte view for block hosts. 714 lines, plus a 780-line doctest suite that links it with **zero** platform code. What it can't be is a media *class*, because in this tree "media class" means "subclass of `MediaType`" — and there's no such thing as *the* `MediaType` to subclass.

Each platform directory declares its own: 13 separate `class MediaType`s, eleven of them under the identical `_MEDIA_TYPE_` header guard, so most pairs can't even be included in one translation unit. `media.h` is just an `#ifdef BUILD_*` hub that picks one. And the same-named virtuals don't agree:

```cpp
virtual error_is_true read(uint32_t sectornum, uint32_t *readcount) = 0;          // rs232
virtual error_is_true read(uint16_t sectornum, uint16_t *readcount) = 0;          // atari
virtual error_is_true read(uint32_t blockNum, uint16_t *count, uint8_t *b) = 0;   // apple
virtual bool          read(uint32_t blockNum, uint8_t *buffer) = 0;               // mac
```

`mount()` differs too (h89 takes a raw `FILE*` and a `mediatype_t`; rs232 adds `fujiHost*`/`filename`), `status()` returns four bytes, one byte, or a bool depending on platform, and `enum mediatype_t` itself is redefined per platform with disjoint enumerators. There is no signature a single top-level class could override.

So the options were: make the format engine subclass rs232's `MediaType` and weld it to one bus — the next platform would have to re-extract it, ending up exactly at this split — or keep the engine standalone and let each platform add only glue. And the adapter mostly isn't forwarding: of `diskTypeIMD.cpp`'s 273 lines, `sector_size()` is the only pure pass-through. The rest is RS232's own contract — the PERCOM block, the 4-byte drive status, controller-status bit mapping, the `DISK_SECTORBUF_SIZE` mount guard. An H89/RC2014/S100 adapter would be a similar-sized file expressing *its* bus's contract, with zero new IMD knowledge. (Those are follow-ups — the h89 and s100 media layers currently don't compile at all, a pre-existing issue.)

The thing that would make it "ready to use" in the plug-in sense is one shared `MediaType` base across all 13 platforms. Worth wanting — but that's a repo-wide refactor, not this PR. Meanwhile 12e18bc9c adds the why to `IMDImage.h`'s header comment so it's in the code, not just this thread.

deltecent:
Heads-up on a small collision with #1582 (`MediaTypeFloppy`, RS232 `.DSK`). Both PRs append a new value to the wire-visible `mediatype_t` in `lib/media/rs232/diskType.h` right after `MEDIATYPE_ROM`, so as written they both take value **3**.

Since that enum is emitted as the mediatype byte in directory entries (append-only, as your comment notes), I propose we lock in a fixed order so neither of us renumbers the other:

- `MEDIATYPE_IMD` = 3 (this PR, keep as-is)
- `MEDIATYPE_FLOPPY` = 4 (#1582 defers to 4)

I'll rebase #1582 onto 4 whenever this lands — no change needed on your side. The rest of the overlap (`media.h` include, `discover_mediatype()`, the `mount()` switch, and the CMake/test files) is complementary and merges cleanly.

Nice touch on the `#ifndef FNIO_IS_STDIO` guard in `fnio.h`, by the way — #1582's `diskfloppy_tests` use the same `FNIO_IS_STDIO` + `UNIT_TESTS` isolation, so that formalizes exactly what our tests rely on.

PR #1588, opened 2026-08-30
fujiDevice doesn't need its own bootdisk virtualDevice

FozzTexx:
Removed `fujiDevice::bootdisk`. It was an extra virtual device that wasn't really part of the bus/daisyChain. Apple II and Adam had already proven that using disk drive 0 worked fine and didn't need an extra virtualDevice.


tschak909 (approved):
Yeah, that's fine.

PR #1590, opened 2026-08-31
Add A8CAS "fsk " chunk playback to Atari cassette (FUJI path)

calaluis:
## Summary

Adds standards-based reproduction of A8CAS raw FSK chunks (`Chunk_Type == "fsk "`,
fourth byte `0x20`) to the normal FUJI cassette playback path in the Atari SIO
cassette subsystem (`lib/device/sio/cassette.cpp`, guarded by `BUILD_ATARI`).

Previously, `send_FUJI_tape_block` recognized only `data` and `baud` chunks; any
other chunk, including `fsk `, was silently skipped by advancing past it, so any
signal carried by an `fsk ` record was dropped. This change reproduces `fsk `
chunks per the A8CAS CAS format specification, treating `fsk ` as one chunk
type that may be interleaved with `FUJI`, `baud`, and `data` in a single image.

The implementation is deliberately **generic and standards-based**. Detection is
purely by the 4-byte chunk type. There is no filename-, game-, country-, or
title-specific logic anywhere in the change.

## What changed

- **New pure module `lib/device/sio/fsk_plan.{h,cpp}`** — hardware-independent,
  allocation-free, I/O-free step functions plus a small O(1) cursor. Handles:
  - `fsk ` chunk parsing with **little-endian** decoding of `chunk_length`, the
    `aux`/IRG field, and each 16-bit FSK signal value.
  - Duration decoding in A8CAS units of **1/10 ms**, converted to **1 MHz RMT
    ticks** (`value × 100`, exact — no rounding).
  - Logical level from **original signal-index parity** (even → logical 0, odd →
    logical 1), independent of preceding durations.
  - **Zero-duration values** consume a parity index but emit no portion, so the
    logical level of every following value is preserved.
  - Splitting long durations into consecutive same-level **portions of at most
    32767 ticks** (the RMT 15-bit field limit) on the fly, with O(1) carried
    state and no materialized waveform.
- **`play_fsk_chunk` (cross-platform)** — owns one FSK chunk end to end:
  bounds-clamping, a **single pre-read** of the clamped payload into a heap
  buffer *before* any signal emission (no file I/O during emission), IRG
  handling, signal reproduction (ESP) or safe skip (PC), and a single idempotent
  cleanup path. Never calls `setBaudrate`.
- **ESP32 RMT raw signal output** — a stateful RMT simple encoder
  (`fsk_encode_cb`, `IRAM_ATTR`) drives `PIN_UART2_TX` at 1 MHz in one continuous
  `rmt_transmit`, mirroring the existing Turbo 2000 encoder precedent. On
  teardown it reattaches UART2 TX routing, leaving the **active baud divisor
  untouched** so `data` records after an FSK record transmit at the correct rate.
- **Safe PC-build behavior** — where `ESP_PLATFORM` is undefined, no RMT/GPIO is
  used; the IRG is honored deterministically via `SYSTEM_BUS.bus_idle`, bounds
  and pre-read logic are identical, and playback continues without raw signal
  generation.
- **Malformed / truncated / odd-length handling** — all reads are bounded by the
  smaller of the declared length and bytes remaining in the image; no
  out-of-bounds reads. Odd lengths ignore the trailing unpaired byte; truncated
  headers and overruns terminate safely at the real boundary (end-of-tape).
- **Chunk-walk integration** — `send_FUJI_tape_block` gains an `fsk ` branch
  before the unknown-chunk catch-all and continues walking after a handled FSK
  chunk (FSK is not a terminating record). Turbo 2000 and QROS paths are
  unchanged; QROS continues to skip `fsk ` exactly as before.
- **Tests** — new host-buildable `fsk_plan_tests` (doctest), registered in
  `tests/CMakeLists.txt`, covering the A8CAS worked example plus generated-loop
  property checks over the value/length space.

Existing formats are preserved: a `baud`/`data`-only image plays back exactly as
before, and only the FUJI path is modified.

## Verification

Automated and build verification:

- **`fsk_plan_tests`: pass** — 14/14 test cases, ~13.77M assertions, exit 0.
- **fujinet-pc build: pass** — shared FSK code compiles and links cleanly with no
  ESP-only RMT/GPIO APIs leaking into the PC build.
- **Full `fujinet-atari-v1` ESP32 firmware build: pass** — built via the official
  `build.sh` in the configured PlatformIO environment; `firmware.elf` linked and
  `firmware.bin` image generated successfully.

Manual real-hardware acceptance:

- A real-world CAS image, `Night Knight.cas`, was used **only as a local
  manual hardware-compatibility example**. It loaded and played back correctly
  through interleaved `baud`, `data`, and `fsk ` chunks, with `data` after FSK
  continuing at the active baud (including the 600 → 790 baud transition).
  **This file is NOT included in this repository** — it is not committed, not
  added as a test fixture, and not redistributed. All behavior is derived solely
  from chunk contents, never from the filename or any fixture-specific value.
- FSK IRG behavior was hardware-tested using a **local synthetic modified
  fixture** (not committed) to exercise inter-record gap timing.

## Known limitation

The live **motor-line de-assertion abort during an FSK IRG longer than 1000 ms**
has **not** been physically exercised on hardware. The abort/recovery code path
(safe abort, UART restoration, baud preservation, retry from the record's
starting offset) was verified by **code inspection only**, which passed. This PR
does **not** claim full hardware verification of that specific abort trigger.

## Test plan

- [x] `fsk_plan_tests` passes on the host.
- [x] fujinet-pc builds cleanly.
- [x] Full `fujinet-atari-v1` ESP32 firmware builds (ELF + BIN generated).
- [x] Manual: `Night Knight.cas` loads and plays on real Atari + FujiNet hardware
      (local only; file not committed).
- [x] Manual: FSK IRG timing exercised with a local synthetic modified fixture.
- [ ] Motor-line abort during a >1000 ms FSK IRG — not physically exercised;
      code-inspection-reviewed only.

### DIFF
--- a/.kiro/specs/a8cas-fsk-chunk-playback/.config.kiro
+++ b/.kiro/specs/a8cas-fsk-chunk-playback/.config.kiro
@@ -1,0 +1,1 @@
+{"specId": "8fbe1e1e-2344-4e02-ba9b-d1d1aa0cbd7a", "workflowType": "requirements-first", "specType": "feature"}
--- a/.kiro/specs/a8cas-fsk-chunk-playback/design.md
+++ b/.kiro/specs/a8cas-fsk-chunk-playback/design.md
@@ -1,0 +1,1002 @@
+# Design Document: A8CAS FSK Chunk Playback
+
+## Overview
+
+This design adds reproduction of A8CAS raw FSK chunks (`Chunk_Type == "fsk "`, fourth byte 0x20 SPACE) to the normal FUJI cassette playback path in `lib/device/sio/cassette.cpp` (guarded by `BUILD_ATARI`). Today `send_FUJI_tape_block` walks the chunk stream recognizing only `data` and `baud`; any other chunk (including `fsk `) is silently skipped by advancing the read offset past it. As a result, images that interleave `fsk ` records with normal `baud`/`data` records (e.g. the Chilean commercial title "Night Knight.cas") lose the signal carried by the FSK records.
+
+The change inserts `fsk ` handling into the existing chunk-walk loop of the FUJI_Playback_Path only. It honors the Inter-Record Gap carried in the chunk `aux`/`irg_length` field, then reproduces the FSK signal on the SIO Data_Line. The design is organized around three pieces:
+
+1. **A pure, hardware-independent set of step functions** (`fsk_plan`) that decode individual little-endian values, derive a logical level from index parity, split a duration into RMT-sized portions, and advance a small memory-bounded cursor over the raw FSK bytes. These functions touch no file, no RMT, no GPIO, and no FujiNet globals, so they are unit-testable on the host, and they do NOT materialize the whole waveform. The decode/parity/scale/split RULES are `static inline`, IRAM-safe arithmetic in the header so the production ISR callback can inline the exact same rules it is tested against; the cursor step (`fsk_view_step`) is a host-test-only helper that shares those same rules but is never called from the ISR.
+2. **A single pre-read step** that reads the entire clamped FSK data region into a heap buffer in ONE `fnio::fread` *before* any hardware emission begins, so the emitted waveform never depends on SD/TNFS/network latency.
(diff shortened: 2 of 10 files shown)


calaluis:
Closing this PR voluntarily after further validation work.

The `Night Knight.cas` image used as the real-hardware acceptance example does not correspond to the original Turbo Software loader implementation, so it is not a valid reference for verifying Turbo Software compatibility.

The original Turbo Software loader source is available here:
https://github.com/a8dogdark/atari_turbo_software

Because the hardware-validation basis used for this PR is therefore not reliable enough for the intended compatibility claim, I am withdrawing the change rather than leave a PR whose acceptance evidence is uncertain.

The generic A8CAS `fsk ` implementation may still be technically useful, but it should be re-evaluated and re-tested against CAS images produced from, or demonstrably matching, the original Turbo Software source before being proposed again.

PR #1594, opened 2026-09-01
New DaisyChain class to centralize device chain management across buses

FozzTexx:
Move the common device chain handling out of the individual bus implementations and into new DaisyChain class. This eliminates the duplicated device ordering, ID management, swapping, rotation, and lookup logic that each bus previously maintained independently, while leaving bus-specific behavior in the bus implementations.


FozzTexx:
IMG_3726
IMG_3729
IMG_3734
IMG_3735
IMG_3744

tschak909 (approved):

PR #1599, opened 2026-09-02
[bus][adam] make adamnet directory reads much faster.

tschak909:


FozzTexx:
Need a test that exhibits the problem on existing firmware and will show it's fixed on this PR, so I can run it on real hardware and make sure that it doesn't break anything. As I've found out **repeatedly** the hard way: what works on the emulator does not always work on real hardware. Especially anything mucking with the bus protocol.

Perhaps @RichStephens can help with making a test.

tschak909:
How do you do that for performance issues? This isn't a PASS/FAIL.

tschak909:
The problem in action. 

https://github.com/user-attachments/assets/90240c18-612d-4d8a-853e-baa7963a6eb2

FozzTexx:
> The problem in action.

Not slow like that on real hardware. 

https://github.com/user-attachments/assets/29efbb8d-2200-48c2-9123-d4462af4f7dc

tschak909:
I'm closing this, as I've been able to make the requisite change in adamcore itself.

PR #1600, opened 2026-09-02
Rename ProtocolParser to NetworkProtocolFactory

FozzTexx:
Ground work for N: Device 1108


tschak909 (approved):
yeah, that makes sense.

PR #1601, opened 2026-09-03
O2 fujinet bringup

tschak909:
The state of O2 bringup
* Config works, can load roms
* 5 Card Stud works, but man is it spartan.


tschak909:
@FozzTexx I think this'll be stable for a while.

FozzTexx (dismissed):

tschak909:
ok, I've brought this up to date.

RichStephens (approved):

PR #1603, opened 2026-09-03
Don't blindly try to strip N: prefix from URL that doesn't have it

FozzTexx:
By the time a URL is being handled by the N: device, the unit number has already been determined. There's no need for a URL to be sent with the N: prefix. Don't blindly look for the first ':' and assume that it is the N: prefix.


tschak909 (commented):
that's fine.

tschak909 (approved):

PR #1605, opened 2026-09-05
[pico][astrocade] FujiNet for the Bally Astrocade

tschak909:
An RP2040 cartridge serving the Astrocade's 8K cart window, and through it a FujiNet: browse a TNFS host, boot over the network, keep talking after boot.

**Emulation only. No cartridge hardware exists yet.**

The cart edge has A0–A12, D0–D7 and one Enable for reads in `0x2000–0x3FFF` — no `/RD`, `/WR`, clock or reset, and writes below `0x4000` go to the magic function generator. So the cart is electrically a ROM and both directions ride the read path: console→cart via hotspot reads where A0–A7 is the payload byte, cart→console via bytes core0 paints into the window. `firmware/include/fuji_mailbox.h` is the spec, shared by the firmware, the MAME device and the Z80 clients. Protocol v2 adds GAME banking (256K/512K, byte-compatible with MAME's `rom_256k`/`rom_512k`) and APPBANK (mailbox stays live).

Also: the `fujiversal-astrocade` board and pinmap (S3 is just the USB host, same as `fujiversal-o2`; **pin numbers provisional**), and CI jobs for the host tests, Z80 clients and cart firmware. `tools/zmac` is vendored, public domain, provenance in its `PROVENANCE.md`.

The ROM push arrives in 512-byte chunks like every other `push_stream()` consumer. The cart spills them to flash in 4K sectors, and erasing one before ACKing the chunk that filled it put the reply outside the ESP32's 100ms read window — so `fuji_store_idle()` blanks the store ahead of time from the mailbox loop and a chunk costs a program alone. #1667 widened the window instead and is closed.

### Testing
Rebuilt on `master` `209b0e0e1` rather than rebased: the old branch replayed the merged O2 series plus an intellivision fix master already has, and had accidentally deleted `data/webui/config/fujiversal-o2.yaml` in a bad merge.

`make atari-lwm` 11/11 ctest. `fujiversal-astrocade` builds. Host tests (`test_fujibus`, `test_astromap`, `test_bankserve`, `test_fujimail`, `test_fujistore`) pass under `-Wall -Wextra -Werror`. Z80 clients assemble and validate. In MAME: `fujicfg` browses and boots over the network, `abcheck.lua` A/Bs against MAME's own mappers, `soak.sh` 133/133.

**No hardware.** Nothing flashed; no cartridge exists. The pinmap is unverified against a board.


tschak909:
Ok Claude. This appeared in the O2 bring-up too. 

There should be ZERO reason to mess with anything on the ESP32 side in lib. Get rid of it, and if something needs to be dealt with, deal with it on the Pico side. You are trying to expose internal bus stuff to the outside which is a BIG NO NO.

tschak909:
Follow-up to the #1667 discussion, since that turned out to be the wrong end to fix this: bb0c5f0b5 moves the flash erase out of the push path.

`fuji_store_idle()` blanks the store a sector at a time from the mailbox service loop on any pass where the console sent nothing, so a chunk that fills the 4K sector buffer pays for a program (~11 ms) rather than erase+program, and the ACK lands inside the existing 100 ms read window with nothing changed on the firmware side. Three guards: no sweep while a push is open, none while the flash store is being served, none once it is blank. A mount that beats the sweep still erases in path — unchanged behaviour, not a regression.

One real bug fell out of it: the busy rule only looked at `fuji_live`. Between the close that stages an image and the console's armed swap read the image is not being served yet, so a second push could have erased it out from under the staged pointers. It checks `fuji_next` now too.

New host test `test_fujistore.c`, with a stub that models the part rather than faking it — a program only clears bits, so a wrongly skipped erase corrupts the image instead of quietly passing. Ten mutations of the bookkeeping, each killed.

Still no hardware.

FozzTexx (approved):

FozzTexx:
> moves the flash erase out of the push path

What flash erase? This cartridge doesn't even exist yet as far as I'm aware, and when it does, wouldn't it be the standard of RP2350 controls the bus and the ROM lives in the RAM of the RP2350?

PR #1606, opened 2026-09-05
Index the day-of-week tables with % 7, not & 7

RichStephens:
DOW3 and DOW2 hold seven entries, so a masked weekday of 7 reads one past the end of both. Use the remainder, which cannot produce 7.


tschak909 (approved):
Yeah. That makes sense. Good catch.

PR #1610, opened 2026-09-06
[all] fix heap leaks on error paths

tschak909:
Part 2 of the heap-depletion audit series (#1610–#1613). Rebased onto current `master` (past #1616 and #1624) and now a single commit — #1609 has merged, so the diagnostics commit that used to ride along here is gone.

**Stands alone against `master`.** Single commit, no dependency on #1611, #1612, #1613 — review and merge it in any order.

_One coordination note:_ this PR changes `iwm_ll.cpp:1137` from `xQueueCreate(10, …)` to a named depth macro, and #1612 adds a null check immediately under that same line. Different lines, so they should merge in either order; if git does complain, that one spot is why.

These are genuine leaks — allocations that escape on error paths and deplete the heap over hours of runtime:

- **`fnFsHTTP::cache_file()`**: the `fc_handle` (plus its `FileHandlerMem`) leaked on all four error exits, including the common GET>399/404 path — small internal-DRAM allocations lost on every failed HTTP file open.
- **CP/M `RAM` (64K)**: the iwm/coco `cpmTask` re-malloc'd RAM on every `while(1)` CCP restart with no free; the atari/rs232/rc2014 `init_cpm` leaked the previous RAM on repeated `CPM_INIT` and left `RAM` dangling after the BIOS-0 `free()` (latent double-free). RAM is now allocated once and reused, and allocation failure is reported instead of `memset(NULL, …)`.
- **Disk II write ISR** (`iwm_ll.cpp`): when the 10-deep write queue was full, `xQueueSendFromISR`'s failure was ignored — the item was dropped and its multi-KB buffer leaked, unfreeable from ISR context. Now the ISR checks for queue space *before* allocating (race-free: it is the sole producer) and counts drops, which the consumer reports. The decode path also frees the item instead of writing through a NULL `malloc` result.
- **fnConfig load**: `FILE*` leaked when the ini-buffer malloc failed; the FLASH path also dereferenced a NULL `FILE*` if the open failed.

**Dropped from the original version of this PR:** the drivewire network `write()` `txbuf` leak. #1624 replaced `drivewireNetwork::write()` with `NDevice::fujidev_write()`, which receives into a `ByteBuffer` (`std::vector<uint8_t>`) instead of a raw `malloc`, so the not-connected exit can no longer leak. Nothing to port — the fix is obsolete. `txbuf` no longer appears anywhere under `lib/device`.

Verification on the rebased branch: `fujiapple-rev0`, `fujiversal-drivewire`, `fujinet-atari-v1`, `fujinet-rs232-rev0` and `fujimac-rev0` firmware builds all pass, as do fujinet-pc COCO and APPLE (4/4 tests each). `fujinet-rc2014spi-rev0` fails, but it fails **identically at `959f6d8f5` (master) with this branch not applied** — `NDevice.cpp` referencing `transaction_accept`/`transaction_error`/`transaction_success` and `FUJI_COMMAND_PACKET`, none of which the rc2014 `systemBus` provides. Pre-existing #1624 fallout, unrelated to this PR. (`fujimac-rev0`, which was broken when this PR was first opened, builds again since #1633.)

Runtime check: 50 failed HTTP file opens should hold internal heap flat in the 10s telemetry.

https://claude.ai/code/session_014RVdfsiC2b8Cc24j45qbCh


FozzTexx:

FozzTexx:
Your intern screwed up this PR and opened it against the wrong branch. I'm guessing this is supposed to go in after #1609 and layer onto that one. Your intern needed to wait until #1609 was merged before opening this one.

RichStephens (dismissed):

RichStephens:
Yeah.... Maybe @FozzTexx is gonna have to fix the conflicts on this one.

tschak909:
Rebased onto master — both of the things you flagged are handled.

@FozzTexx — the hold is honored: this no longer touches `lib/device/*/network.{h,cpp}` at all. The only conflict was a modify/delete on `lib/device/drivewire/network.cpp`, and the hunk it carried (a missing `free(txbuf)` on the not-connected exit of `drivewireNetwork::write()`) is obsolete after #1624 — `NDevice::fujidev_write()` receives into a `ByteBuffer`, so there's no raw allocation left to leak. Dropped rather than ported; nothing needed re-targeting into `NDevice.cpp`.

The wrong-branch problem is also gone: now that #1609 has merged, the diagnostics commit dropped itself on rebase and this PR is a single commit against master.

One thing worth knowing separately from this PR: `fujinet-rc2014spi-rev0` does not build at master (`959f6d8f5`) — `NDevice.cpp` calls `transaction_accept`/`transaction_error`/`transaction_success` and takes a `FUJI_COMMAND_PACKET`, none of which the rc2014 `systemBus` has. Confirmed by building the env at master with this branch unapplied; it fails identically there. Every other env I built is green (`fujiapple-rev0`, `fujiversal-drivewire`, `fujinet-atari-v1`, `fujinet-rs232-rev0`, `fujimac-rev0`, plus fujinet-pc COCO/APPLE).

lib/device/rc2014/rc2014cpm.cpp
```diff
@@ -52,20 +53,29 @@ void rc2014CPM::rc2014_handle_cpm()
     {
         cpmActive = false;
         free(RAM);
+        RAM = NULL;
     }
 }
 
-void rc2014CPM::init_cpm(int baud)
+bool rc2014CPM::init_cpm(int baud)
```
FozzTexx: Don't use `bool`, use `error_is_true` or `success_is_true`. `bool` doesn't tell anyone which polarity it is. On the check use `.is_error()` or `.is_success()`.

lib/device/rs232/rs232cpm.cpp
```diff
@@ -46,22 +47,31 @@ void rs232CPM::rs232_handle_cpm()
     {
         cpmActive = false;
         free(RAM);
+        RAM = NULL;
     }
 }
 
-void rs232CPM::init_cpm(int baud)
+bool rs232CPM::init_cpm(int baud)
```
FozzTexx: Same here, `error_is_true` or `success_is_true`.

lib/device/sio/siocpm.cpp
```diff
@@ -52,20 +53,29 @@ void sioCPM::sio_handle_cpm()
     {
         cpmActive = false;
         free(RAM);
+        RAM = NULL;
     }
 }
 
-void sioCPM::init_cpm(int baud)
+bool sioCPM::init_cpm(int baud)
```
FozzTexx: This one too, `error_is_true` or `success_is_true`.

tschak909:

lib/device/rc2014/rc2014cpm.cpp
tschak909: Done — `init_cpm()` now returns `success_is_true`, returns via `RETURN_ERROR_AS_FALSE()` / `RETURN_SUCCESS_AS_TRUE()`, and the call site reads `cpmActive = init_cpm(...).is_success();`.

lib/device/rs232/rs232cpm.cpp
tschak909: Done — `init_cpm()` now returns `success_is_true`, returns via `RETURN_ERROR_AS_FALSE()` / `RETURN_SUCCESS_AS_TRUE()`, and the call site reads `cpmActive = init_cpm(...).is_success();`.

lib/device/sio/siocpm.cpp
tschak909: Done — `init_cpm()` now returns `success_is_true`, returns via `RETURN_ERROR_AS_FALSE()` / `RETURN_SUCCESS_AS_TRUE()`, and the call site reads `cpmActive = init_cpm(...).is_success();`.

tschak909:
Review notes applied — all three were the same point, so all three sites got the same treatment.

`init_cpm()` in `siocpm.cpp`, `rs232cpm.cpp` and `rc2014cpm.cpp` now returns `success_is_true` instead of a bare `bool`, using the house macros:

```c
success_is_true sioCPM::init_cpm(int baud)
{
    ...
    if (RAM == NULL)
    {
        Debug_printv("could not allocate 64K CP/M RAM, free heap: %lu", fnSystem.get_free_heap_size());
        RETURN_ERROR_AS_FALSE();
    }
    ...
    RETURN_SUCCESS_AS_TRUE();
}
```

and the call sites now read the polarity out explicitly:

```c
cpmActive = init_cpm(9600).is_success();
```

Each header picks up an explicit `#include "global_types.h"` (it already came in transitively through `bus.h`, but `sio.h`/`rs232.h`/`rc2014bus.h` all include it directly, so this matches).

Those three `init_cpm()` declarations were the only bare `bool` this PR introduced — the rest of the diff adds no new return values.

Rebased and amended into the single commit, so the PR is still one commit against master.

**Builds:** `fujinet-atari-v1` and `fujinet-rs232-rev0` both green. `fujinet-rc2014spi-rev0` still fails, unchanged: the errors are all in `rc2014bus.cpp` (`fujiDeviceID_t` vs `uint8_t` in the device map) and the `FUJI_COMMAND_PACKET` mixin headers, and the compiler never reaches `rc2014cpm.cpp`. I compiled that one file standalone with the env's own flags, with and without this branch applied — identical output both ways, so nothing here contributes to it.

FozzTexx (approved):

PR #1611, opened 2026-09-06
[all] fix allocator mismatch and buffer overruns

tschak909:
Memory corruption found while auditing allocations. Part 3 of the heap series; stands alone against `master`.

- **ATX allocator mismatch**: sector data and sector lists were `heap_caps_malloc`'d under ESP but freed everywhere with `delete[]`, including the `AtxTrack` destructor. Both platforms now use `new (std::nothrow)`, with a null check instead of `fread`ing into null.
- **meatloaf flash/tnfs `obtain()`**: `new char[m_path.length()]` left no room for the NUL and `strncpy` never terminated it, so the `strchr` mkdir walk read past the allocation.
- **sio modem**: the firmware buffer was used one line before its null check.

### Testing
`make atari-lwm` 11/11 ctest. `./build.sh -b` for `fujinet-atari-v1` and `fujinet-iec-lolin-d32`, both SUCCESS.

**Not flashed, no hardware verification, no test coverage.** Nothing exercises ATX parsing or the meatloaf path walk; the allocator change is reasoned from every `delete[]` site.


FozzTexx:
Your intern screwed up this PR and opened it against the wrong branch. I'm guessing this is supposed to go in after #1610 and layer onto that one. Your intern needed to wait until #1610 was merged before opening this one.

tschak909:
Rebased. This is now a single commit against `heap-02-leaks` (#1610), which is itself a single commit against current master — the stale duplicates of #1609 and #1610 that made this PR show three commits and conflict are gone.

Nothing in this one needed re-aiming: it touches ATX, meatloaf and the sio modem, none of which #1616/#1624 moved. Verified at master that all three bugs are still live: the `heap_caps_malloc`/`delete[]` mismatch (5 release sites, `diskTypeAtx.cpp:519,530,539,580,589,608`), `new char[m_path.length()]` + `strncpy` with no room for the NUL (`flash.cpp:462`, `tnfs.cpp:413`), and the sio modem NULL check one line after the pointer is used (`modem.cpp:257`).

FozzTexx:

tschak909:

lib/meatloaf/device/flash.cpp
```diff
@@ -459,8 +459,8 @@ void FlashHandle::obtain(std::string m_path, std::string mode) {
         // For file creation, silently make subdirs as needed.  If any fail,
         // it will be caught by the real file open later on
 
-        char *pathStr = new char[m_path.length()];
-        strncpy(pathStr, m_path.data(), m_path.length());
+        char *pathStr = new char[m_path.length() + 1];
+        strcpy(pathStr, m_path.c_str());
```
FozzTexx: Can we just do this the right way and use a std::string instead of a `char *`?

lib/meatloaf/network/tnfs.cpp
```diff
@@ -410,8 +410,8 @@ void TNFSHandle::obtain(std::string m_path, std::string mode) {
         // For file creation, silently make subdirs as needed.  If any fail,
         // it will be caught by the real file open later on
 
-        char *pathStr = new char[m_path.length()];
-        strncpy(pathStr, m_path.data(), m_path.length());
+        char *pathStr = new char[m_path.length() + 1];
+        strcpy(pathStr, m_path.c_str());
```
FozzTexx: Same here, use a std::string

lib/meatloaf/device/flash.cpp
tschak909: Done. The manual buffer is gone in both places — the mkdir walk now runs straight off the std::string:

```cpp
for (size_t pos = m_path.find('/'); pos != std::string::npos; pos = m_path.find('/', pos + 1))
    mkdir(m_path.substr(0, pos).c_str(), ALLPERMS);
```

Same mkdir sequence as before for every path shape I checked (relative, absolute, trailing slash, doubled slashes, no slash), minus the allocation and the temporary NUL-poking.

lib/meatloaf/network/tnfs.cpp
tschak909: Same change here — tnfs.cpp and flash.cpp were byte-identical in this block and both now walk the std::string with find()/substr(), no char* copy at all.

lib/media/atari/diskTypeAtx.cpp
```diff
@@ -526,11 +527,13 @@ success_is_true MediaTypeATX::_load_atx_chunk_sector_data(chunk_header_t &chunk_
...
     
     // Attempt to the sector data
-#ifdef ESP_PLATFORM
-    track.data = (uint8_t *)heap_caps_malloc(data_size * sizeof(uint8_t), MALLOC_CAP_DEFAULT);
-#else
-    track.data = new uint8_t[data_size];
-#endif
+    // Every free of this buffer is delete[], so it must come from new[].
+    track.data = new (std::nothrow) uint8_t[data_size];
+    if (track.data == nullptr)
```
FozzTexx: From `CONTRIBUTING.md`:

> Own every resource and check every allocation, including on error and teardown paths. Prefer std::unique_ptr or a container to a bare new/delete pair. Heap leaks on error paths are a recurring bug class here.

lib/media/atari/diskTypeAtx.cpp
```diff
@@ -576,11 +579,12 @@ success_is_true MediaTypeATX::_load_atx_chunk_sector_list(chunk_header_t &chunk_
     }
 
     // Attempt to read sector_header * sector_count
-#ifdef ESP_PLATFORM
-    sector_header_t *sector_list = (sector_header_t *)heap_caps_malloc(track.sector_count * sizeof(sector_header_t), MALLOC_CAP_DEFAULT);
-#else
-    sector_header_t *sector_list = new sector_header_t[track.sector_count];
-#endif
+    sector_header_t *sector_list = new (std::nothrow) sector_header_t[track.sector_count];
```
FozzTexx: From `CONTRIBUTING.md`:

> Own every resource and check every allocation, including on error and teardown paths. Prefer std::unique_ptr or a container to a bare new/delete pair. Heap leaks on error paths are a recurring bug class here.

RichStephens: Is this resolved to your satisfaction? Looks like it to me, unless I'm misunderstanding what's going on.

FozzTexx: It's still doing `new`

tschak909: Fixing it.

tschak909: Fixed in 3b74ccb5e — the sector list is now a `std::unique_ptr<sector_header_t[]>`, so both `delete[]`s are gone (including the one on the short-read path).

The allocation is still `new (std::nothrow)` rather than a `std::vector` on purpose: `CONFIG_COMPILER_CXX_EXCEPTIONS` is off in every ESP sdkconfig, so `vector::resize` failing would abort the firmware instead of returning. `sector_count` comes straight out of the image file, so a corrupt ATX needs to fail the load, not reset the board. `unique_ptr` gives the ownership CONTRIBUTING asks for while keeping the checked allocation.

lib/media/atari/diskTypeAtx.cpp
tschak909: Same fix in 3b74ccb5e: `AtxTrack::data` is now a `std::unique_ptr<uint8_t[]>`, so the reload path, the short-read path and the destructor all stop freeing by hand — `~AtxTrack()` is deleted outright.

That has a second effect worth calling out: the user-declared destructor was suppressing `AtxTrack`'s implicit move constructor, so any `_tracks` reallocation would have copied the raw `data` pointer and double-freed it. It only survived because the constructor reserves all 40 tracks up front.

Allocation stays `new (std::nothrow)` with the null check rather than a `std::vector` — exceptions are off in the ESP builds (`CONFIG_COMPILER_CXX_EXCEPTIONS` unset everywhere) and `data_size` comes from the chunk header, so a bad ATX has to fail the load instead of aborting.

`make atari-lwm` 11/11, `./build.sh -b` for `fujinet-atari-v1` SUCCESS.

lib/media/atari/diskTypeAtx.cpp
```diff
@@ -554,7 +546,7 @@ success_is_true MediaTypeATX::_load_atx_chunk_sector_data(chunk_header_t &chunk_
     // Keep a count of how many bytes we've read into the Track Record
     track.record_bytes_read += data_size;
 
-    //util_dump_bytes(track.data, 64);
+    //util_dump_bytes(track.data.get(), 64);
```
FozzTexx: From CONTRIBUTING.md:

> Do not leave commented-out code. Delete it; git has it.

tschak909: Gone in b08324d0c. It predates this PR but I touched the line, so it's mine now — deleted rather than updated.

I swept the rest of the PR diff for the same thing; that was the only commented-out call I'd added or modified.

FozzTexx (approved):

PR #1612, opened 2026-09-06
[all] check FreeRTOS task, queue and semaphore creation for failure

tschak909:
Task stacks, queues and semaphores come from internal DRAM (`pvPortMalloc` is hard-wired to `MALLOC_CAP_INTERNAL`), so they fail first under pressure. 15 task-create and a dozen queue sites never checked the result. Each now follows the #1608 pattern: check, null the handle, log free internal/total heap, then a per-site action.

| Site | On failure |
|---|---|
| `fnLoop`, `adamnet_bus` | `abort()` — panic-reboot instead of a booted-looking brick |
| adamnet RESET / card-detect | don't install an ISR whose queue didn't create; `xQueueSendFromISR` on null crashes in ISR context |
| adamnet printer | `thPrinter` was uninitialized, so a failed create made the destructor `vTaskDelete()` garbage |
| iwm / drivewire / N: CP/M | stop reporting a live session when queues or task never started |
| Web UI reset | fall back to `esp_restart()` |
| ACMChannel | real checks instead of `assert()`, which compiles out under NDEBUG |
| Disk II write queue | gate the ISR install on the queue existing |

Split per review: 17 files → 13. #1662 esp_timer, #1663 dead spiMutex, #1664 empty-vector OOB, #1665 CP/M transaction assert, #1666 url-decode null check.

Task and queue checks stayed together — in `iwm/cpm.cpp`, `drivewire/cpm.cpp`, `CPM.cpp`, `printer.cpp` and `fnSystem.cpp` they're the same conditional, so splitting means landing half-guarded code then rewriting it. Say the word if you want them apart anyway.

### Testing
Single commit against `master`. `make atari-lwm` 11/11 ctest. `./build.sh -b` for `fujiapple-rev0`, `fujinet-adam-v1`, `fujinet-coco-devkitc`, `fujinet-atari-v1` — all SUCCESS. Full `-a` on the pre-split change: 26 built, 5 failed, all 5 failing identically on clean `master` (#1658, #1659).

**Not flashed, no hardware verification, no test coverage.** These paths only run under real heap exhaustion. Every site is reasoned from the ESP-IDF contract, not observed failing. The two `abort()` sites are the judgement call worth a second opinion.

Known-unchecked, left alone: `fnWiFi.cpp` event groups, `adamnet/serial.cpp` queue.


FozzTexx:

FozzTexx:
Your intern screwed up this PR and opened it against the wrong branch. I'm guessing this is supposed to go in after #1611 and layer onto that one. Your intern needed to wait until #1611 was merged before opening this one.

tschak909:
Rebased onto the refreshed #1611; single commit again, current base.

One thing changed in the rebase, @FozzTexx: the two `esp_timer_create` checks this carried for `sioNetwork::timer_start()` and `rs232Network::timer_start()` are **dropped**, not re-aimed. #1624 replaced the rate timers with `NDevice::poll_interrupt()`, which is software timing with no `esp_timer` behind it — `timer_start`/`onTimer` have no definition anywhere in the tree now. So this PR no longer touches `lib/device/*/network.{h,cpp}` at all. The commit message and the table above now say ×3 (TNFS keep-alive, LED flicker, ATX rotation) instead of ×5.

Everything else re-applied unchanged and is still live at master — spot-checked: `fnLoop` (`main.cpp:657`) and `adamnet_bus` (`adamnet.cpp:505`) unchecked, the adamnet RESET queue unchecked while the ISR is installed regardless, `thPrinter` still an uninitialized member (so the existing destructor null-guard tests garbage), `rxQueue` in `ACMChannel::begin()` with no check of any kind while the rest use `assert()`, the dead `spiMutex`, and the Disk II write ISR still calling `xQueueSendFromISR` on an uninitialised handle (`iwm_ll.cpp:921`, declared without an initialiser at `iwm_ll.h:294`).

Incidental note for whoever owns those platforms: `H89Network.h:180`, `rc2014Network.h:189` and `s100spiNetwork.h:198` still declare `rateTimerHandle` / `timer_start()` / `timer_stop()` with no definition and every call site commented out. Left alone here — dead code, but not this PR's job.

FozzTexx:

tschak909:

FozzTexx:
> If a change cannot be explained in a couple of sentences, it is probably more than one change. Split it into a series of small branches that each open their own pull request against master

tschak909:
Fair. Split into five against `master`: #1662 esp_timer checks, #1663 dead spiMutex, #1664 empty-vector OOB writes, #1665 CP/M success tripping `transaction_error()`'s assert, #1666 url-decode null check (network-reachable). This PR is 17 files down to 13.

One I did not split: the task-create and queue-create checks. In `iwm/cpm.cpp`, `drivewire/cpm.cpp`, `CPM.cpp`, `printer.cpp` and `fnSystem.cpp` they are the same conditional, so separating them lands half-guarded code in one PR and rewrites it in the next. Say so and I will do it anyway.

Also: `./build.sh -a` turned up five boards that do not build on clean `master`, unrelated — #1658 and #1659. None are in the CI matrix.

Force-pushed.

FozzTexx (dismissed):
I know this PR is just doing a bunch of checking to make sure pointers aren't NULL, but since it touched adamnet.cpp I needed to find the time to test it on real hardware. I have been burned **too many times** with how overly sensitive the Adam is to changes in adamnet.cpp and I wanted to make sure that it would still boot, run CONFIG, and could run one of the tests.

Good news: Adam still works!

RichStephens (dismissed):
Looks good to me, once the conflicts are resolved.

tschak909:
Rebased onto master. One conflict, in `ACMChannel::begin()`: #1682 moved the `usb_host_install` + `usb_lib` task-create out to `usbHostEnsureInstalled()` in `fnUsbHost.cpp`, carrying the `assert(task_created == pdTRUE)` with it verbatim. Took master's call site and moved this PR's hardening to where the code now lives — so the abort-on-failure is in `fnUsbHost.cpp` instead.

Rebuilt after the rebase: `fujiversal-intv` (the board that actually compiles `ACMChannel.cpp`/`fnUsbHost.cpp`), plus `fujinet-adam-v1`, `fujinet-atari-v1`, `fujinet-coco-devkitc`, `fujiapple-rev0` — all SUCCESS. `make atari-lwm` 11/11.

FozzTexx (approved):

PR #1613, opened 2026-09-06
[apple2] move the modem queues to PSRAM

tschak909:
Two `xQueueCreate(32770, 1)` FIFOs in the `iwmModem` constructor are ~65.7K of internal DRAM at boot on a plain ESP32 — close to the whole free internal heap after WiFi, and likely the largest fixed cost behind the depletion #1608 chased. `pvPortMalloc` is hard-wired to `MALLOC_CAP_INTERNAL`, so `CONFIG_SPIRAM_MALLOC_ALWAYSINTERNAL` never reached them.

Both are task-to-task byte FIFOs with no ISR access (verified: no `FromISR` use), so `xQueueCreateWithCaps(…, MALLOC_CAP_SPIRAM)` moves them to PSRAM unchanged. Mac modem is the same, ~32.9K. Creation checked, destructor pairs with `vQueueDeleteWithCaps`, handles get in-class `= nullptr`.

Split per review: 13 files → 4. #1660 the `shrink_to_fit` removals, #1661 WOZ1 scratch, #1664 the empty-vector OOB write.

### Testing
Single commit against `master`. `make atari-lwm` 11/11 ctest. `./build.sh -b` for `fujiapple-rev0` and `fujimac-rev0`, both SUCCESS.

**Not flashed, no hardware verification.** The 65.7K is arithmetic on the queue geometry, not a measurement. Cheap check, no bench needed: boot a DEBUG `fujiapple-rev0` and read the first 10s telemetry line — free internal DRAM should be up ~65K vs `master`. Nobody has. PSRAM queue access is slower than internal; unmeasured for modem throughput.


FozzTexx:

FozzTexx:
Your intern screwed up this PR and opened it against the wrong branch. I'm guessing this is supposed to go in after #1612 and layer onto that one. Your intern needed to wait until #1612 was merged before opening this one.

tschak909:
Rebased onto the refreshed #1612; single commit, current base.

Two rebase decisions worth flagging:

- The `shrink_to_fit()` removals for `drivewireNetwork::read()` and `sioNetwork::sio_read()` **collapsed into one site**. #1624 unified both read paths into `NParser::read`, which is non-virtual and is what every platform now funnels through (`fujidev_read` → `fujicore_read` → `NParser::read`), so the one deletion at `NParser.cpp:12` does what the two used to, for more buses than before.
- The `iecNetwork::transmit()` shrink is **dropped** — the refactor deleted that path outright (`iecNetwork.cpp` is 12 lines now, and `NParser::write` has no shrink).

Net count went 9 → 11: while re-aiming I picked up four more per-transfer sites of exactly the same shape that the original audit had skipped — `QRMixin.cpp:100`, the `CLIPBOARD.cpp` write path, and the two legacy `rc2014Fuji.cpp` base64 copies. Close-time and one-shot shrinks are still untouched (`Protocol.cpp:133-135`, `CLIPBOARD::close`, `fnDirCache`, `fnFsSD`, S3/GDRIVE/ONEDRIVE `close_file_handle`/`umount`).

Everything else re-applied unchanged. `xQueueCreateWithCaps` still appears nowhere else in the tree, and all the targets are still live at master: both modem queue pairs (`iwm/modem.cpp:98` 2×32770, `mac/modem.cpp:145` 2×16384, all four handles uninitialised with unconditional `vTaskDelete`/`vQueueDelete` in the destructors), the 6.6K `MALLOC_CAP_INTERNAL` WOZ1 scratch used only to skip 6 bytes per track, and both empty-`std::vector` `operator[]` writes.

Verification, since it can't be seen from the diff: `./build.sh -p ATARI` builds fujinet-pc clean and all 9 host test suites pass, and the stack tip builds green on the ESP envs that cover this code.

FozzTexx:
Your intern just doesn't get it. It's still a PR against the wrong branch, even if it rebases it.

FozzTexx:

tschak909:

FozzTexx:
Can we do all the shrink_to_fit removals in their own PR?

tschak909:
Done — #1660, just the 11 `shrink_to_fit` removals.

This PR is now only the modem queues moving to PSRAM: 13 files down to 4. WOZ1 scratch went to #1661, the empty-vector OOB write to #1664.

Force-pushed, so the diff is smaller than when you last read it.

FozzTexx (approved):

PR #1620, opened 2026-09-08
Consolidate webui

RichStephens:
Use a base UI with only platform files (that inherit from the base) when they differ from the base UI.  Anything that doesn't have a specific platform file gets the plain base UI.

Ran a confirmation script that built the UIs for all platforms vs master.  They showed no differences except where we made tweaks to fix some omissions from the original or UI pieces that didn't need to be there that were in the original.


tschak909:
what happened?

PR #1621, opened 2026-09-08
Consolidate webui configs with inheritance

### LINKED ISSUE #1517: Consolidate `data/webui` configuration
The `data/webui` directory currently contains a separate config YAML file for each board. There are dozens of these files, even though most of their contents are identical.

This should be consolidated so that there is **one base/template configuration per bus**, with only small board-specific additions or overrides where necessary.

Benefits:
- Reduce the number of nearly-identical files.
- Changes to the common configuration only need to be made in one place.
- Make it easier for new boards to inherit the standard configuration.
- Reduce the chance of configurations getting out of sync.

There should only need to be a small handful of config files rather than one for every board. A small amount of scripting may be needed to generate or assemble the final configurations.

### LINKED ISSUE #1577: [apple2] Web interface doesn't show Disk II slots
The web interface is missing the Disk II slots.

Image

RichStephens:
Consolidate webui YAML configs from 45+ individual board files to base + 9 platforms + 6 PC variants using inheritance and loader fallback.

All 29 firmware boards produce identical output to master. All PC builds verified successful.

Fixes #1623 
Fixes #1577


FozzTexx:
What's in "the PC versions" that requires them to be different?

data/webui/config/BUILD_RS232.yaml
```diff
@@ -0,0 +1,6 @@
+extends: base
+components:
+  hsio_settings: true
```
FozzTexx: hsio is an Atari-only thing

data/webui/config/fujinet-pc-rs232.yaml
```diff
@@ -0,0 +1,5 @@
+extends: BUILD_RS232
+components:
+  udp_stream: true
```
FozzTexx: I think `udp_stream` is gone, replaced by netstream

RichStephens:
@FozzTexx I believe I've addressed all the issues. Here are the actual differences in the FujiNet-pc versus the hardware ones.

## fujinet-pc differences from hardware

**Component overrides (YAML-level):**

| Board | Override |
|---|---|
| fujinet-pc-adam | `serial_port: true` |
| fujinet-pc-apple | *(none)* |
| fujinet-pc-atari | `serial_port: true` |
| fujinet-pc-coco | `serial_port: true` |
| fujinet-pc-rs232 | *(none)* |

**Template-level UI differences (gated by `tweaks.fujinet_pc`):**

Common to all 5 PC boards:
- Page title source: `FN_LABEL` vs `FN_HOSTNAME`
- Network module: "Device name" field vs "Host name" field; second row shows Host name vs IP address; PC adds 3 blank rows (IP address, Default gateway, DNS server)
- Hardware module: different stat set (System/uptime/restart button vs SPIFFS/SD/heap/CPU)
- Mount list: PC adds Mount All/Swap buttons vs a MOUNT ALL DISKS link on hardware
- BOIP module: PC adds an editable hostname/IP field; hardware shows a "reboot required" tooltip instead (justified — `setHost()` only applies live on PC, per `httpServiceConfigurator.cpp:640-644`)

Platform-specific additions:
- **fujinet-pc-apple**: mount slot count is 8 vs 10 on hardware (`tweaks.fujiapple`)
- **fujinet-pc-atari**: HSIO current-setting display/dropdown differs (Atari-only feature)

No longer differences (unified during this audit):
- CoCo Becker Port default-port sentence — now shown on both
- Timezone module header — now "Time Zone" on both

FozzTexx:
> * Page title source: `FN_LABEL` vs `FN_HOSTNAME`

Why can't they be the same?

> * Network module: "Device name" field vs "Host name" field; second row shows Host name vs IP address; PC adds 3 blank rows (IP address, Default gateway, DNS server)

Again, why can't they be the same?

> * Mount list: PC adds Mount All/Swap buttons vs a MOUNT ALL DISKS link on hardware

Again, it seems like they're different just for the sake of being different. I sort of recall maybe @nwah opened an issue for this or was trying to change it so the `mount all` looked the same on both.

> fujinet-pc-apple: mount slot count is 8 vs 10 on hardware (tweaks.fujiapple)

https://github.com/FujiNetWIFI/fujinet-firmware/issues/1577

> fujinet-pc-atari: HSIO current-setting display/dropdown differs (Atari-only feature)

This is a little iffy since LWM can connect to a real Atari via a TTL serial adapter and HSIO would still apply on LWM.

RichStephens:
> > * Page title source: `FN_LABEL` vs `FN_HOSTNAME`
> 
> Why can't they be the same?
> 
> > * Network module: "Device name" field vs "Host name" field; second row shows Host name vs IP address; PC adds 3 blank rows (IP address, Default gateway, DNS server)
> 
> Again, why can't they be the same?
> 
> > * Mount list: PC adds Mount All/Swap buttons vs a MOUNT ALL DISKS link on hardware
> 
> Again, it seems like they're different just for the sake of being different. I sort of recall maybe @nwah opened an issue for this or was trying to change it so the `mount all` looked the same on both.
> 
> > fujinet-pc-apple: mount slot count is 8 vs 10 on hardware (tweaks.fujiapple)
> 
> #1577
> 
> > fujinet-pc-atari: HSIO current-setting display/dropdown differs (Atari-only feature)
> 
> This is a little iffy since LWM can connect to a real Atari via a TTL serial adapter and HSIO would still apply on LWM.


FN_LABEL vs FN_HOSTNAME: 

PC: <%FN_LABEL%> → Config.get_general_label() (PC-only code, doesn't compile on ESP32)
Hardware: <%FN_HOSTNAME%> → fnSystem.Net.get_hostname()

Network module.  Not 100% sure. Checking.  It's different, but not sure why it HAS to be.

Mount All/Swap - Again - not sure why it HAS to be. Maybe has to do with the SWAP button.

Slot count - Yeah... maybe this needs to wait until that gets fixed

HSIO on pc-Atari.  Dunno. Guess I need to look into it.

FozzTexx:
There's still one `FN_HOSTNAME` that needs to be changed:
```
Hacking on xtreme9:: git cb ; git cc ; find data/webui -type f -exec grep -IHn FN_HOSTNAME {} \; 
consolidate-webui
ab72fcf43
data/webui/template/www/index.tmpl.html:115:                                    <div class="det detlinecol"><%FN_HOSTNAME%></div>
```

tschak909 (approved):
Thank you.

PR #1623, opened 2026-09-09
Enable FN_LABEL and FN_DEVICE_NAME on all platforms

RichStephens:
## Summary
Enable FN_LABEL (custom device label with hostname fallback) and FN_DEVICE_NAME template variables on hardware and PC builds, not just PC-only.

The underlying dependencies (devicename config field and fnSystem.Net.get_hostname()) are available on all platforms. This allows the webui to use a unified label source across all platforms instead of platform-specific conditionals in templates.

## Changes
- `lib/config/fnc_general.cpp`: Moved `#ifndef ESP_PLATFORM` guard to wrap only PC-specific helper functions, leaving `get_general_label()` available on all platforms
- `lib/http/httpServiceParser.cpp`: Removed platform guards from FN_DEVICE_NAME and FN_LABEL template variable handlers

### DIFF
--- a/lib/config/fnConfig.h
+++ b/lib/config/fnConfig.h
@@ -108,9 +108,7 @@ public:
 
     // GENERAL
     std::string get_general_devicename() { return _general.devicename; };
-#ifndef ESP_PLATFORM
     std::string get_general_label();
-#endif
     int get_general_hsioindex() { return _general.hsio_index; };
     std::string get_general_timezone() { return _general.timezone.empty() ? "UTC" : _general.timezone; };
     bool get_general_rotation_sounds() { return _general.rotation_sounds; };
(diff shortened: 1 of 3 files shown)


FozzTexx (dismissed):

FozzTexx:
Now I have to ask why keep FN_LABEL at all? If the goal is to make all the web pages the same, then shouldn't it be either FN_LABEL or FN_HOSTNAME?

RichStephens:
> Now I have to ask why keep FN_LABEL at all? If the goal is to make all the web pages the same, then shouldn't it be either FN_LABEL or FN_HOSTNAME?

The pages are all going to have FN_LABEL.  That's why I made this change. I haven't pushed that change to the other PR yet.

FozzTexx (approved):

PR #1626, opened 2026-09-11
Tolerate SD cards that reject CMD59 (CRC_ON_OFF) on the SPI path

### LINKED ISSUE #1625: SD-over-SPI mount fails on cards that reject CMD59 (CRC_ON_OFF): sdmmc_card_init failed (0x106)
### Summary

On the SD-over-SPI code path, some microSD cards fail to mount under ESP-IDF with:

```
E (xxxx) sdmmc_sd: sdmmc_init_spi_crc: sdmmc_send_cmd_crc_on_off returned 0x106
E (xxxx) vfs_fat_sdmmc: sdmmc_card_init failed (0x106).
```

`0x106` is `ESP_ERR_NOT_SUPPORTED`. The same cards mount fine under the Arduino `SD`
library, and often work in other readers, so they are not defective.

This affects the SPI branch in `lib/FileSystem/fnFsSD.cpp` — `FileSystemSDFAT::start()`,
which uses `SDSPI_HOST_DEFAULT()` + `esp_vfs_fat_sdspi_mount()`. The 4-bit SDMMC host
branch is **not** affected (in SD mode, CRC is mandatory and handled in hardware).

### Root cause

ESP-IDF's SD init runs an unconditional step in SPI mode, `sdmmc_init_spi_crc()`, which
sends **CMD59 (CRC_ON_OFF)** to turn on CRC checking. In `components/sdmmc/sdmmc_init.c`:

```c
SDMMC_INIT_STEP(is_spi, sdmmc_init_spi_crc);   /* runs for every SPI card */
```

Some cards answer CMD59 with the R1 *illegal-command* bit set. The sdspi transaction
layer maps that bit to `ESP_ERR_NOT_SUPPORTED` (0x106), `sdmmc_init_spi_crc()` logs the
error and returns it, and `sdmmc_card_init()` aborts the entire mount — so the card never
comes up and file access is unavailable.

CRC16 of data transfers is **optional** in SPI mode per the SD Physical Layer spec
(mandatory only in SD/native mode). The Arduino ESP32 `SD` library's SPI diskio
(`sd_diskio.cpp`) never sends CMD59 at all, which is exactly why these cards work there.

There is no Kconfig option or `sdmmc_host_t` flag in current ESP-IDF to skip that step.

### How we resolved it

We hit this in our own ESP-IDF/SPI project,
[deltecent/fdc-sds-pio](https://github.com/deltecent/fdc-sds-pio) (an ESP32 FDC+ serial
disk server), and made the CRC-enable step tolerant instead of fatal: attempt CMD59, and
if the card rejects it, continue with CRC off (i.e. the Arduino behavior). Cards that
accept CMD59 keep CRC on. The change is in
[`src/disk/sd.c`](https://github.com/deltecent/fdc-sds-pio/blob/master/src/disk/sd.c).

The least invasive way we found is a linker wrap of the IDF init step, so no IDF sources
are patched. It works because `sdmmc_init_spi_crc` is a real (non-static, non-inline)
symbol that the linker can redirect.

**1. Add the linker flag.** With PlatformIO, in `platformio.ini`:

```ini
build_flags = -Wl,--wrap=sdmmc_init_spi_crc
```

With plain ESP-IDF / CMake, add it to your target instead:

```cmake
target_link_options(${COMPONENT_LIB} PUBLIC "-Wl,--wrap=sdmmc_init_spi_crc")
# or, at the project level:
# target_link_options(${PROJECT_NAME}.elf PUBLIC "-Wl,--wrap=sdmmc_init_spi_crc")
```

**2. Provide the wrapper** in any one of your `.c` files that gets linked (e.g. next to
the SD mount code). This is the complete, drop-in definition:

```c
#include "esp_err.h"
#include "esp_log.h"
#include "sdmmc_cmd.h"   // brings in the sdmmc_card_t type

// Exported by the sdmmc component (declared in
// components/sdmmc/include/esp_private/sdmmc_common.h). Forward-declared here so
// we don't need the private header.
esp_err_t sdmmc_send_cmd_crc_on_off(sdmmc_card_t *card, bool crc_enable);

// The linker redirects calls to sdmmc_init_spi_crc() here; the original remains
// reachable as __real_sdmmc_init_spi_crc() if ever needed.
esp_err_t __wrap_sdmmc_init_spi_crc(sdmmc_card_t *card)
{
    // Try to enable CRC (CMD59). We call it directly instead of
    // __real_sdmmc_init_spi_crc() because that wrapper does an ESP_LOGE on
    // failure, printing a "returned 0x106" line on every boot with these cards.
    esp_err_t err = sdmmc_send_cmd_crc_on_off(card, true);
    if (err != ESP_OK) {
        // Card rejected CMD59. CRC16 is optional in SPI mode, so leave it off
        // (the Arduino SD library never enables it either) and let the mount
        // proceed. Cards that accept CMD59 fall through with CRC on.
        ESP_LOGI("sd", "card does not support CMD59/CRC_ON_OFF (%s); "
                       "continuing with SPI CRC off", esp_err_to_name(err));
    }
    return ESP_OK;   // never fatal
}
```

That's the whole change — two files, no IDF sources patched. After it, the affected
cards mount normally over SPI and the console is clean.

(An alternative, if you prefer not to wrap: replicate the mount sequence without the
CRC-enable step, or keep calling `__real_sdmmc_init_spi_crc()` and swallow a non-OK
return — but that still prints the IDF `ESP_LOGE` line each boot.)

### Suggested direction for fujinet-firmware

Since this only affects the SPI SD path, wrapping/tolerating the CMD59 step in the SPI
branch of `FileSystemSDFAT::start()` would let SPI-wired boards accept the broader set of
cards that already work under Arduino and in card readers, without touching the SDMMC path.

### Environment

- ESP32, microSD wired for SPI (`esp_vfs_fat_sdspi_mount`)
- Reproduced with multiple 4 GB FAT32 cards that mount fine under the Arduino `SD` library

tschak909:
Fixes #1625.

## Problem

Some microSD cards answer **CMD59 (CRC_ON_OFF)** with the R1 illegal-command bit set. The sdspi transaction layer maps that to `ESP_ERR_NOT_SUPPORTED` (0x106), ESP-IDF's `sdmmc_init_spi_crc()` step treats it as fatal, and `sdmmc_card_init()` aborts the whole mount:

```
E (xxxx) sdmmc_sd: sdmmc_init_spi_crc: sdmmc_send_cmd_crc_on_off returned 0x106
E (xxxx) vfs_fat_sdmmc: sdmmc_card_init failed (0x106).
```

The cards aren't defective — CRC16 of data transfers is *optional* in SPI mode per the SD Physical Layer spec (mandatory only in SD/native mode), and the same cards mount fine under the Arduino `SD` library, whose SPI diskio never sends CMD59 at all.

## Fix

IDF exposes no Kconfig option or `sdmmc_host_t` flag to skip that init step, so it is redirected with a linker wrap — no IDF sources patched:

- `-Wl,--wrap=sdmmc_init_spi_crc` added to the shared `[env] build_flags` in `platformio-ini-files/platformio.common.ini` (and `platformio-sample.ini` for the VS Code users), so every board picks it up through `${env.build_flags}`.
- `__wrap_sdmmc_init_spi_crc()` next to the mount code in `lib/FileSystem/fnFsSD.cpp`: it attempts CMD59, and if the card rejects it, logs one `Debug_printf` line and continues with CRC off. Cards that accept CMD59 keep CRC on. It never returns an error, so the mount proceeds either way.

It calls `sdmmc_send_cmd_crc_on_off()` directly rather than `__real_sdmmc_init_spi_crc()` because the real step does an `ESP_LOGE` on failure, which would print the alarming "returned 0x106" line on every boot with an affected card.

Only the SPI path runs this step — in SD/4-bit mode CRC is mandatory and handled in hardware — so the `SDMMC_HOST_WIDTH` branch of `FileSystemSDFAT::start()` is unaffected. The wrapper is defined unconditionally on ESP builds so the link is correct regardless of which branch a board uses, and it is inside `#ifdef ESP_PLATFORM`, so FujiNet-PC builds are untouched.

## Verification

Built `fujinet-v1-8mb` (Atari, SPI SD, IDF 5.5) and checked the linked ELF:

- `__wrap_sdmmc_init_spi_crc` is present and the IDF init-step table points at it; the real `sdmmc_init_spi_crc` is gone from the image entirely, i.e. its only reference was redirected.
- Disassembly of the wrapper is exactly `sdmmc_send_cmd_crc_on_off(card, true)` → on error `Debug_printf(...)` → `return 0`.

Not tested against physical hardware here — the reporter has the affected cards. Original analysis and the approach come from @deltecent in #1625.

https://claude.ai/code/session_01K6tJLB5ci4VRQfSrwwAJSu

### DIFF
--- a/lib/FileSystem/fnFsSD.cpp
+++ b/lib/FileSystem/fnFsSD.cpp
@@ -620,6 +620,46 @@ const char * FileSystemSDFAT::partition_type()
 }
 
 #ifdef ESP_PLATFORM
+/*
+  Some SD cards answer CMD59 (CRC_ON_OFF) with the R1 illegal-command bit set, which the
+  sdspi layer reports as ESP_ERR_NOT_SUPPORTED (0x106). ESP-IDF runs that step
+  (sdmmc_init_spi_crc) unconditionally for every card brought up over SPI and treats a
+  failure as fatal, so sdmmc_card_init() aborts and the card never mounts:
+
+    E (xxxx) sdmmc_sd: sdmmc_init_spi_crc: sdmmc_send_cmd_crc_on_off returned 0x106
+    E (xxxx) vfs_fat_sdmmc: sdmmc_card_init failed (0x106).
+
+  CRC16 of data transfers is optional in SPI mode per the SD Physical Layer spec (it is only
+  mandatory in SD/native mode), which is why these cards work under the Arduino SD library -
+  its SPI diskio never sends CMD59 at all. IDF offers no Kconfig option or sdmmc_host_t flag
+  to skip the step, so it is redirected here with "-Wl,--wrap=sdmmc_init_spi_crc" (see the
+  build_flags in platformio-ini-files/platformio.common.ini) and made non-fatal: cards that
+  accept CMD59 keep CRC on, cards that reject it mount with CRC off instead of failing.
+
+  Only the SPI path reaches this step - in SD/4-bit mode CRC is mandatory and handled by the
+  hardware - so the SDMMC branch of start() below is unaffected.
+
+  See https://github.com/FujiNetWIFI/fujinet-firmware/issues/1625
+*/
+extern "C"
+{
+    // Exported by the sdmmc component, declared in its private header
+    // esp_private/sdmmc_common.h - declared here so we don't have to reach into that header.
+    esp_err_t sdmmc_send_cmd_crc_on_off(sdmmc_card_t *card, bool crc_enable);
+
+    esp_err_t __wrap_sdmmc_init_spi_crc(sdmmc_card_t *card)
+    {
+        // CMD59 is attempted directly rather than through __real_sdmmc_init_spi_crc(), which
+        // remains linked, because that one does an ESP_LOGE on failure - printing the alarming
+        // "returned 0x106" line on every boot with an affected card.
+        esp_err_t e = sdmmc_send_cmd_crc_on_off(card, true);
+        if(e != ESP_OK)
+            Debug_printf("SD card rejected CMD59/CRC_ON_OFF (%s), continuing with SPI CRC off\r\n", esp_err_to_name(e));
+
+        return ESP_OK; // never fatal
+    }
+}
+
 success_is_true FileSystemSDFAT::start()
 {
     if(_started)
--- a/platformio-ini-files/platformio.common.ini
+++ b/platformio-ini-files/platformio.common.ini
@@ -52,6 +52,9 @@ build_flags =
     -D ${fujinet.flash_filesystem}
     -D CMAKE_EXPORT_COMPILE_COMMANDS=ON
     -D DEBUG_SPEED=${env.monitor_speed}
+    ; Make the SPI SD CRC-enable step (CMD59) non-fatal - see __wrap_sdmmc_init_spi_crc()
+    ; in lib/FileSystem/fnFsSD.cpp
+    -Wl,--wrap=sdmmc_init_spi_crc
 
 # [fujinet] section allowed values.
 # copy one of the build_board values to your platformio.local.ini under [fujinet], or read the build-sh.md docs.
--- a/platformio-sample.ini
+++ b/platformio-sample.ini
@@ -133,6 +133,9 @@ build_flags =
     -D ${fujinet.flash_filesystem}
     -D CMAKE_EXPORT_COMPILE_COMMANDS=ON
     -D DEBUG_SPEED=${env.monitor_speed}
+    ; Make the SPI SD CRC-enable step (CMD59) non-fatal - see __wrap_sdmmc_init_spi_crc()
+    ; in lib/FileSystem/fnFsSD.cpp
+    -Wl,--wrap=sdmmc_init_spi_crc
     ;-D LEAK_DEBUG           ; Show low heap usage every service loop.
     ;-D NO_BUTTONS          ; if your hardware has no physical buttons
     ;-D JTAG                ; enable use with JTAG debugger


tschak909:
@deltecent can you test this PR and see how it goes? :)

mozzwald:
I tested this patch with 5 different SD cards. My thought was the at least one of them was a card that did not work in FujiNet. I tested them all with an older firmware `FujiNet v1.6.1 2026-06-18 07:17:11 (ATARI)` and then on this branch. Both firmware versions worked same, so at least the patch does not break cards that previously worked.

deltecent:
I have not had time to test, but will on the RS232 this weekend.

These are the cards I have that did not work when I first tried to the RS232 firmware and ran into the same  problem with the new FDC+ firmware: https://www.amazon.com/dp/B0C39YHGC3

RichStephens:
> I have not had time to test, but will on the RS232 this weekend.
> 
> These are the cards I have that did not work when I first tried to the RS232 firmware and ran into the same problem with the new FDC+ firmware: https://www.amazon.com/dp/B0C39YHGC3

@deltecent  Have you had a chance to test this to see if it works with those bad cards?

deltecent:
Sorry, @RichStephens, I got sidetracked on a vi editor for CP/M. I will test tomorrow. Thank you.

deltecent:
Tested on hardware — works.

Built this branch (f8e2d40e0) for `fujinet-rs232-s3` and flashed it to an RS232 Rev1 ESP32S3 with one of the affected microSD cards. Boot log:

```
LittleFS mounted.
SD card rejected CMD59/CRC_ON_OFF (ESP_ERR_NOT_SUPPORTED), continuing with SPI CRC off
SD mounted.
fnConfig::load
Load fnconfig.ini from SD
fopen = /fnconfig.ini rb : ok
fnConfig::load read 921 bytes from config file
```

The card that previously aborted the mount with `sdmmc_card_init failed (0x106)` now mounts cleanly and config loads from it. No alarming `ESP_LOGE` line, just the single informational message.

Claude on behalf of @deltecent

RichStephens (approved):
Tested good, looks good. Approved.

PR #1627, opened 2026-09-12
Add HS Uart support for devkit-c on the coco fujinet

wdathing:
This is in support of using the coco centipede board and the onerom dongle-free upgrade for the coco fujinet.


data/webui/config/fujinet-coco-devkitc-centipede.yaml
```diff
@@ -0,0 +1,24 @@
...
+  net_stream: false
+  program_recorder: experimental
+  disk_swap: false
+  emulator_settings: true
+  boot_settings: true
+  cpm_settings: experimental
+  pclink: false
+tweaks:
+  # webui tweaks, if any
+  platform: COCO
```
FozzTexx, line 24: @RichStephens just changed things in PR #1621 so that it isn't necessary add yaml files for every board type.

lib/device/drivewire/drivewireFuji.cpp
```diff
@@ -198,6 +202,61 @@ void drivewireFuji::random()
...
+// the DBC exchange itself completed) before this returns, so the caller sees
+// no interleaving with the DBC side channel.
+//
+// Selecting a file in the config app's browser only records its path (see
+// FUJI_SET_DEVICE_FULLPATH) - it does not itself mount anything, and nothing
+// else mounts device slots ahead of a normal DECB cold boot (FUJI_MOUNT_ALL).
+// A "!LDCART!" boot can be triggered directly from the browser/config app
+// without ever going through that boot path, so this mounts the slot itself
+// (read-only - a cartridge image is inherently read-only anyway) rather than
+// assuming it already happened. Idempotent if it's already mounted.
```
FozzTexx, line 218: I'm pretty sure @RichStephens and @nwah already have things setup so that after the ROM image is sent it doesn't automatically swap in on the pico on close. There's a bit in the control register so the CoCo can choose when to enable the ROM image.

lib/hardware/ESP32UARTChannel.cpp
```diff
@@ -126,7 +140,7 @@ void ESP32UARTChannel::updateFIFO()
         {
             size_t old_len = _fifo.size();
             _fifo.resize(old_len + event.size);
-            int result = uart_read_bytes(_uart_num, &_fifo[old_len], event.size, 0);
+            int result = uart_read_bytes(_uart_num, &_fifo[old_len], event.size, 1);
```
FozzTexx, line 143: This will add a delay which will break Adam.

lib/hardware/ESP32UARTChannel.cpp
```diff
@@ -144,7 +158,7 @@ void ESP32UARTChannel::updateFIFO()
 
     size_t old_len = _fifo.size();
     _fifo.resize(old_len + avail);
-    int result = uart_read_bytes(_uart_num, &_fifo[old_len], avail, 0);
+    int result = uart_read_bytes(_uart_num, &_fifo[old_len], avail, 1);
```
FozzTexx, line 161: This will add a delay which will break Adam.

FozzTexx (commented):
This is too big of a PR, need to break this up into smaller pieces and bring them in one at a time. Separate out the IO/UART changes into a PR, the media changes into a PR, the drivewire bus changes into a PR, changing of the autorun.dsk into a PR, and the new board into a PR.

lib/device/drivewire/drivewireFuji.cpp
wdathing: This one needs a message as we don't have a control register!  Going over the onerom control bus.

lib/hardware/ESP32UARTChannel.cpp
wdathing: Will fix

lib/hardware/ESP32UARTChannel.cpp
wdathing: will fix

data/webui/config/fujinet-coco-devkitc-centipede.yaml
wdathing: I will check that out and fix.

wdathing:
I didn't realize this branch was more than one off master.  I'll withdraw
this request once i have dealt with the other comments that have come in,
but i will put it into draft first.

lib/device/drivewire/drivewireFuji.cpp
FozzTexx: There are 4 registers in the fujiversal coco pico firmware: 
```
.define public IO_BASE          0xFF41
.define public IO_GETC          1
.define public IO_STATUS        0
.define public IO_PUTC          3
.define public IO_CONTROL       2
```
STATUS bit 6 (0x40) will report when the ROM has finished loading into the pico RAM, CONTROL bit 2 (0x04) will switch the ROM in.

PR #1629, opened 2026-09-13
Add DBC push_stream support for MEDIATYPE_ROM images

wdathing:
the centipede/onerom build has a different method for loading roms over to the pico as we dont have register access.  Instead, the coco sends a message to our pico via access a sequence of address that causes the onerom to request the mounted rom image itself.  That's just how this architecture has to work.

### DIFF
--- a/lib/media/drivewire/mediaTypeROM.cpp
+++ b/lib/media/drivewire/mediaTypeROM.cpp
@@ -33,6 +33,70 @@ uint8_t MediaTypeROM::status()
     return 2;
 }
 
+#if defined(PINMAP_FUJIVERSAL_DRIVEWIRE) || defined(COCO_HS_UART)
+#define DBC_STREAM_ROM 0
+
+// Pushes this object's already-mounted file (_media_fileh, _media_image_size)
+// to the DBC device (the RP2040/RP2350 companion) as one
+// NET_OPEN/NET_WRITE.../NET_CLOSE stream. Shared by mount() (automatic push,
+// for boards wired that way) and fujiDevice's on-demand FUJI_PULL_ROM handler
+// (push triggered later, e.g. by a One ROM-style companion's own knock/ack
+// handshake) - see lib/media/rs232/diskTypeROM.cpp's push_stream() for the
+// sibling implementation this mirrors.
+bool MediaTypeROM::push_stream()
+{
+    struct { uint8_t id; u32le_t size; } open_hdr;
+    static_assert(sizeof(open_hdr) == 5, "OPEN header must not be padded");
+    open_hdr.id = DBC_STREAM_ROM;
+    open_hdr.size = _media_image_size;
+
+    auto reply = SYSTEM_BUS.sendCommand(FUJI_DEVICEID::DBC, CMD::NET_OPEN,
+                                        std::string((const char *)&open_hdr, sizeof(open_hdr)));
+    if (!reply || reply->command() != CMD::FUJI_ACK)
+    {
+        Debug_printv("MediaTypeROM: failed to open DBC stream (%lu bytes)", (unsigned long)_media_image_size);
+        return false;
+    }
+
+    fnio::fseek(_media_fileh, 0, SEEK_SET);
+
+    bool ok = true;
+    uint32_t sent = 0;
+    size_t got;
+    while ((got = fnio::fread(_media_blockbuff, 1, MEDIA_BLOCK_SIZE, _media_fileh)) > 0)
+    {
+        reply = SYSTEM_BUS.sendCommand(FUJI_DEVICEID::DBC, CMD::NET_WRITE,
+                                       std::string((char *)_media_blockbuff, got));
+        if (!reply || reply->command() != CMD::FUJI_ACK)
+        {
+            Debug_printv("MediaTypeROM: failed to send DBC block at %lu of %lu bytes",
+                         (unsigned long)sent, (unsigned long)_media_image_size);
+            ok = false;
+            break;
+        }
+        sent += got;
+    }
+
+    if (ok && sent != _media_image_size)
+    {
+        Debug_printv("MediaTypeROM: DBC stream short transfer: %lu of %lu bytes",
+                     (unsigned long)sent, (unsigned long)_media_image_size);
+        ok = false;
+    }
+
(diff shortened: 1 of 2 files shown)


lib/media/drivewire/mediaTypeROM.cpp
```diff
@@ -33,6 +33,70 @@ uint8_t MediaTypeROM::status()
...
+    open_hdr.id = DBC_STREAM_ROM;
+    open_hdr.size = _media_image_size;
+
+    auto reply = SYSTEM_BUS.sendCommand(FUJI_DEVICEID::DBC, CMD::NET_OPEN,
+                                        std::string((const char *)&open_hdr, sizeof(open_hdr)));
+    if (!reply || reply->command() != CMD::FUJI_ACK)
+    {
+        Debug_printv("MediaTypeROM: failed to open DBC stream (%lu bytes)", (unsigned long)_media_image_size);
+        return false;
+    }
```
FozzTexx, line 59: The media class shouldn't be sending bus commands directly. The media layer should deal with the media itself—reading, writing, opening, closing, getting its size, etc. `SYSTEM_BUS.sendCommand()` is a bus/protocol operation and should be handled by the code that is handling the bus request.

For example, `MediaTypeROM` shouldn't need to know that opening the media requires sending `CMD::NET_OPEN` to the DBC, or even that a DBC exists. It should just provide the media operations that its caller needs. The bus-handling code should translate the bus command into those media operations.

Otherwise we're making the media implementation dependent on the bus protocol, when the whole point of having a media layer is that it represents the media independently of how that media is being accessed.

FozzTexx:
> the centipede/onerom build has a different method for loading roms over to the pico as we dont have register access. 

If there's no registers then how does reading and writing a byte over drivewire work?

wdathing:
>> If there's no registers then how does reading and writing a byte over drivewire work?

So the onerom sits in a rom socket, theres no write line brought to it.  Instead, the host (coco here) reads from a sequence of addresses that are not going to occur at random, then the onerom goes into command mode. For a dw write, we do the sequence (called a knock by the one-rom plugin architecture) then we send the 16 bit length of the write, then the bytes.  the bytes are sent by the host on the address bus as the low order bytes of the address, clocked in on the read strobe.  This methodology has been around since the 80s, used on the smartwatch no slot clock that went into the rom sockets on various early PCs and 8 bit machines.  More recently, this was used by Mike Brutman on the pc jr's cart slot SD card reader in very much the same way i'm using it here.  
For our fujinet purposes, we have 3 different knocks/operations: dwwrite, dwread, and cart load.  Cart load triggers the onerom to send the new DW command to transfer the cart rom via the same sideband channel we're using for fujiversal.

Oh, to finish painting that picture:  this is all running on a standard coco fujinet cart.  You replace the ROM with a onerom Fire28C or Fire28D (both know as Fire28 RGB, differing only in USB connector orientation) rom emulator, based (retail cost 10$) and use 2 jumpers or a cable assembly from the Fire28 over to the dongle connector on the cart.  I load this version of firmware and now we're loading software at 921kbaud.  As opposed to the coco2 at 56kbaud, so that's a nice upgrade for little $$, and we're not landfilling a plastic case, a pcb and the devkitc board.

This code is working, and working pretty well, on the coco2.  I demo'd it on the CocoNation show yesterday.

Also, there's another more generic rp2350 board that uses the devkit c board, same esp32 load, for the coco.  It's called the centipede and gives fujinet, debug capabilties, screen mirroring etc for the coco.

FozzTexx:
I'm guessing you have a heavily modified HDB-DOS ROM to make all that knocking work?

Even though the CoCo filters out writes to the cartridge ROM space, the OneROM should still be able to see when the CoCo is accessing the IO space from 0xFF00 to 0xFFFF. So even though the R/W line isn't connected, you could use four IO bytes instead of just two. This is how it was done on a different bus through an EPROM cartridge and a PicoROM. I'm sure that code is still in the fujiversal repo.

wdathing:
Yes, I modified HDBdos to do this.
No, can't see those addresses.  Coco cart doesn't map A15 or A14, those are hooked up to dip switches.  I am working in a context of a plugin architecture offered by the onerom, and it's all about transfering bytes, so my plugin can only read or write bytes after that knock sequence.  The bus interface is controlled by pio engines and I essential have fifos in and out.
And BTW, i'm not ignoring your other comment about not sending bus commands from here.  I understand the need for abstraction and seperation.  Just need time to think about it, and daytime is doing errands around the house, and watching the dodgers game, so i won't get to that until this evening!

FozzTexx:
> Coco cart doesn't map A15 or A14

Hmm... A 16k cart is from 0xC000-0xFEFF, with the IO space at 0xFF00-0xFFFF, so the top 256 bytes of the ROM space would be the IO space. But without something telling you that it's the high address space via either SCS or A15+A14 I suppose there's no way to know. 🤔

Which EPROM PCB are you using with the OneROM?

wdathing:
It’s the coco fujinet board, rev 000 and rev 0000.

wdathing:
There's a picture of the situation on here:
https://github.com/FujiNetWIFI/fujinet-hardware/tree/master/Coco/One-ROM-Fire-28C

lib/media/drivewire/mediaTypeROM.cpp
wdathing: Actually, my change made it so for my new case, that we *don't* push from the media type.  It's the fujiversal case that does, my change just moves the worker into a new funciton.
Thoughts?

FozzTexx:
@RichStephens put together a PR which gets all the SYSTEM_BUS stuff out of lib/media. https://github.com/FujiNetWIFI/fujinet-firmware/pull/1646

Before we merge PR 1646, how will that affect what you're working on?

wdathing:
Rich's PR takes priority over this one... get that merged and if it breaks this then i'll withdraw this one and adapt to the new way.

FozzTexx:
> Rich's PR takes priority over this one... get that merged and if it breaks this then i'll withdraw this one and adapt to the new way.

Thanks! The other PR is merged, which as expected, caused conflicts with this one.

wdathing:
I'll pull down the new master and make my stuff work with that.  I'll have the replacement stuff up in a few days max.
Thanks,
Bill

PR #1630, opened 2026-09-13
Add FUJI_PULL_ROM command and COCO_HS_UART drivewire bus support

wdathing:
Protocol changes for onerom/centipede, little bit of cleanup

### DIFF
--- a/lib/device/drivewire/disk.h
+++ b/lib/device/drivewire/disk.h
@@ -20,6 +20,7 @@ public:
     ~drivewireDisk();
 
     mediatype_t disktype() { return _media == nullptr ? MEDIATYPE_UNKNOWN : _media->_mediatype; };
+    MediaType *media() { return _media; }
     mediatype_t mount(fnFile *f, const char *filename, uint32_t disksize,
                       disk_access_flags_t access_mode,
                       mediatype_t disk_type = MEDIATYPE_UNKNOWN);
--- a/lib/device/drivewire/drivewireFuji.cpp
+++ b/lib/device/drivewire/drivewireFuji.cpp
@@ -3,6 +3,10 @@
 #include "drivewireFuji.h"
 #include "NDevice.h"
 #include "compat_string.h"
+#include "endianness.h"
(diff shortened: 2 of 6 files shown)


lib/device/drivewire/disk.h
```diff
@@ -20,6 +20,7 @@ class drivewireDisk : public drivewireDevice
     ~drivewireDisk();
 
     mediatype_t disktype() { return _media == nullptr ? MEDIATYPE_UNKNOWN : _media->_mediatype; };
+    MediaType *media() { return _media; }
```
FozzTexx, line 23: This should not be here. `_media` is private because it belongs to the disk and nothing should be accessing it directly.

lib/device/drivewire/drivewireFuji.cpp
```diff
@@ -198,6 +202,61 @@ void drivewireFuji::random()
...
+        return;
+    }
+
+    SYSTEM_BUS.transaction_success();
+#else
+    Debug_println("PULL ROM: not supported on this FujiNet hardware");
+    SYSTEM_BUS.transaction_error();
+#endif
+}
+
```
FozzTexx, line 259: This doesn't belong here. The disk device is the thing that handles the ROM, not the fuji device and the fuji device shouldn't be handling disk related commands.

The disk drive should be able to send the ROM over immediately on a mount anyway because the way it works currently with the Fujiversal pico firmware is the ROM is stuffed into a RAM buffer on the pico and not activated.

What you're really looking for is a way to tell the pico itself to switch the pointer over to the ROM on demand.

wdathing: This isn't where i tell my pico to switch over.  This is in response to my asking the pico to switch over; the pico itself is sending this to pull down the rom over it's uart.
I send a message to the pico; it sends this request, handles the data transfer, does the switchover, then returns when the rom is switched over.  Coco is polling for the status until that happens.
I'll take a look at this further later tonight, as it's still $dayjob time.

PR #1634, opened 2026-09-14
[sdkconfig] re-enable bounded TCP out-of-order queueing (#1615)

### LINKED ISSUE #1615: Intermittent stream buffering or freezing while using telecommunication applications, introduced after v1.6.1
I've been chasing an intermittent buffering/freezing issue in data flow to the terminal, and isolated it to a real regression somewhere after the v1.6.1 release. Reporting the test matrix in case it's useful - I haven't pinned down a root cause myself, just the conditions under which it does and doesn't appear.

Symptom: during an active terminal session, the screen completely freezes for anywhere from several seconds up to a couple of minutes. Keystrokes typed during the freeze still register once it ends (they aren't lost), which suggests something is blocked/stalled rather than truly hung. Once it releases, any queued output appears all at once, then things return to normal - until it happens again, seemingly at random. It's happened over a local WiFi LAN as well as longer connections, so it doesn't appear to be about connection distance alone.

**Test matrix**

1. FujiNet-PC on Windows 11 (fw v1.6-468c9198f 2026-07-27) hosting a BBS in Altirra, connected to via SyncTERM - clean
2. FujiNet-PC on a Mac (fw v1.6-283af1943 2026-08-31) hosting a BBS in Atari800MacX, connected to via MuffinTerm - clean
3. Real Atari 800XL + real FujiNet (firmware: [1.6.2 230 and after]) hosting the BBS, connected to via SyncTERM (from Windows) - buffering/ freezing issue
4. Same real hardware, connected to via MuffinTerm (from Mac) - buffering/freezing issue
5. Real Atari 800XL + real FujiNet, running BobTerm as the terminal client, calling OUT to the Mac's FujiNet-PC-hosted BBS - buffering/ freezing issue
6. Real Atari 800XL + real FujiNet, rolled back to firmware v1.6.1 (last official release), same test as #3/#4 (local BBS use only, no TNFS) - clean, no buffering at all

**Key observations**

- Every combination involving only emulated hardware (Altirra or Atari800MacX + FujiNet-PC, on either Windows or Mac, in either direction) is completely clean.
- The issue only appears when the real physical Atari + real FujiNet device is anywhere in the session - whether it's hosting the BBS, or just running a terminal client (BobTerm) connecting out to an emulated host.
- The issue shows up under two completely separate pieces of terminal/BBS software (a custom BBS I've written, and BobTerm) - which is why I don't believe this is a bug in either of those programs, and instead something at the FujiNet firmware/hardware level.
- Rolling the real hardware back to v1.6.1 (the last official release) removes the issue entirely. This narrows it to a regression introduced somewhere in the nightly builds after v1.6.1 - present by at least v1.6-468c9198f (2026-07-27).

**Expected behavior**

Screen output should flow continuously without multi-second/multi-minute stalls, matching the behavior seen on v1.6.1 and on FujiNet-PC.

**Actual behavior**

Output periodically freezes for seconds to a couple of minutes at a time, then dumps out in a burst, on real hardware running any nightly build tested so far (but not on v1.6.1).

**Additional context**

1. HSIO is disabled on the real hardware.
2. I haven't yet captured a debug log from the exact moment a freeze happens. Happy to grab one from the Flasher's serial console, or try to narrow down which specific nightly introduced this, if that would help.

tschak909:
Fixes #1615.

## Cause

Terminal/BBS sessions freeze for seconds to minutes on real hardware, then dump output in a burst. Every FujiNet-PC combination in the reporter's matrix is clean.


tschak909:
@FozzTexx @RichStephens tested on my eclaire atari800. The modem is operating normally now.

tschak909:
It's worth noting, this particular bug only happens on ESP32 hardware. Not on emulation.

RichStephens (approved):

PR #1636, opened 2026-09-15
[apple2] restore the Disk ][ slots in the web interface

### LINKED ISSUE #1577: [apple2] Web interface doesn't show Disk II slots

tschak909:
Fixes #1577.

## What was wrong

Apple2 has 10 mount slots — 8 SmartPort + 2 Disk ][ — and `MAX_MOUNT_SLOTS` is 10 for *every* `BUILD_APPLE` build, not just the ESP32 one (`lib/config/fnConfig.h:10-15`, no `ESP_PLATFORM` condition).

#1244 nevertheless split the mount list into a non-PC branch with 10 rows and a FujiNet-PC branch with 8, and added `&& defined(ESP_PLATFORM)` to the `FN_DRIVE9`/`FN_DRIVE10` guards in `httpServiceParser.cpp`. FujiNet-PC Apple has shown 8 slots ever since. ESP32 was never affected, which is why #1621 didn't fix it.

Every other slot enumerator — `render_slotpicker`, `mount_file`, `eject_slot`, the REST drives API, `fnconfig` — was already `MAX_MOUNT_SLOTS`-driven and correct on PC. The index page was the last thing pretending there were 8.

**Caveat:** on FujiNet-PC the two restored rows map to `_fnDisks[8..9]`; `DEV_RELAY_SLIP` compiles out `iwmDisk2`, so they display and accept mounts but SLIP never serves them as woz drives. That gap predates this PR and already applied to the slot picker and the REST API.

## Also fixed, same files

- **Disk ][ label off-by-one.** Rows read "Drive 0"/"Drive 1" — #1244 bumped the subtrahend by 5 instead of 4 when the loop grew from 6 slots to 10. Back to 1-based, matching `IWM_ACTIVE_DISK2` and the slot picker.
- **Orphaned HSIO markup (#1621).** An `{% if tweaks.fujinet_pc %}` was removed from inside the HSIO `<select>` but its `{% endif %}` was left behind, so it closed `components.hsio_settings` early and leaked a stray `</select>`, a phantom Save button and a surplus `</div></form>` onto every platform *without* HSIO settings — Apple, Adam, CoCo, IEC, Lynx, RC2014, RS232, Mac, H89. The "Current Setting" value the same commit blanked is restored too.
- **Dead Mount All / Swap buttons (#1621).** Both render on ESP32 now, but neither JS helper was defined there and `/swap` had no route, so they threw `ReferenceError`. The helpers are no longer PC-only, and ESP32 gets a `/swap` handler calling `fujicmd_image_rotate()` from the shared `fujiDevice` base.
- **Missing ESP32 IP address row (#1621).** The `{% if tweaks.fujinet_pc %}Host name{% else %}IP address{% endif %}` pair in the NETWORK module was collapsed to an unconditional "Host name", so no ESP32 build since has rendered `FN_IPADDRESS`/`FN_IPMASK` — both still live in the parser, they just lost their only reference. Restored ahead of Default gateway, with the alt striping flipped behind it. ESP32 now shows Device name, Host name *and* IP address; the first six rows match the PC branch.
- **`pio run -a dev`** replaced the board config with `dev.yaml` outright, so a dev preview of an Apple board lost `tweaks.fujiapple` and rendered the generic 8 "Drive Slot" rows. `dev.yaml` is an overlay now.

## Verification

- `fujiapple-rev0` builds and links for ESP32.
- `httpServiceParser.cpp`, `mgHttpService.cpp`, `httpServiceBrowse.cpp`, `httpServiceApi.cpp` compile clean under `-DBUILD_APPLE -DDEV_RELAY_SLIP`. Tag-table parity confirmed: 110 tags in both the enum and `tagids[]` on PC and ESP alike.
- WebUI rendered for ESP/PC Apple, Atari and Adam. Both Apple variants emit 10 `data-mountslot` rows ending in `Disk ][ Drive 1`/`Disk ][ Drive 2`, and the mount-list block is byte-identical between them. `<select>`/`<form>` counts went from 2/3 and 12/13 to balanced on Apple and Adam; Atari is unchanged apart from the restored HSIO value.

## Out of scope

One thing noticed but left alone, worth a separate issue:

- Atari's program-recorder module has an unmatched `</form>` — present on master, unrelated to this change.


FozzTexx:
> Dead Mount All / Swap buttons (https://github.com/FujiNetWIFI/fujinet-firmware/pull/1621). Both render on ESP32 now,

They do not. On ESP32 it's just "mount all" as text. The mount/swap buttons are still only on LWM.
image

tschak909:
I think what you're looking at is a stale web UI rather than stale code — the firmware on that board is this branch, but `/www` isn't.

On ESP32 the web UI isn't part of the firmware binary. `platformio.ini` sets `data_dir = ./data/${fujinet.build_platform}`, `build_webui.py` renders into it as a `pre:` script, and that tree becomes the LittleFS image. Only `pio run -t uploadfs` (`./build.sh -f`, `make uploadfs`) puts it on the device — `build.sh -u` writes firmware alone and leaves the old `/www` in place. On LWM there's no separate step: the same generated tree is served straight out of the build dir, which is exactly why the buttons are there and not on the ESP32.

Two things in your screenshot date that page to before #1621, never mind this PR:

- The `MOUNT ALL DISKS` link at the top of the mount list. #1621 (d25c72aba) deleted that markup — `git show d25c72aba^:data/webui/template/www/index.tmpl.html`, line 305. Neither master nor this branch can emit it.
- The `IP address   10.4.0.194 / 255.255.255.0` row. #1621 collapsed `{% if tweaks.fujinet_pc %}Host name{% else %}IP address{% endif %}` into an unconditional "Host name", so nothing built since Sep 9 renders `<%FN_IPADDRESS%> / <%FN_IPMASK%>` on ESP32 at all.

Meanwhile the banner reads `v1.6-d0687c15d 2026-09-15 06:42:14` — firmware from this branch, filesystem from before Sep 9. (The `Disk ][ Drive 0`/`Drive 1` labels are the pre-fix ones too.)

After a `./build.sh -f` you should get a `Drive Slots  [Mount All] [Swap]` row at the bottom of the mount list, with both actually wired up: `swapButton()`/`mountAllButton()` are now emitted on every platform (`data/webui/template/www/js/utils.tmpl.js`), `/swap` is registered on ESP32 (`lib/http/httpService.cpp:2202`), and `/mount?mountall=1` was already handled (`httpService.cpp:642`). If it still looks like your screenshot after that, then it *is* a real bug and I'll chase it — please say so.

Your screenshot earned its keep twice over, though:

- That IP address row is a genuine #1621 regression. I'd flagged it as out of scope; since you'd have lost it the moment you re-flashed, I've restored it here instead (62587d69c). ESP32 now shows Device name, Host name *and* IP address.
- `SPIFFS size 0 / SPIFFS used 0` in your Hardware panel is a real bug as well: `FileSystemLittleFS::total_bytes()`/`used_bytes()` call `esp_littlefs_info(NULL, …)`, but the partition is registered with `.partition_label = "storage"`, so the query matches nothing and both return 0. `fnFsSPIFFS.cpp` has the identical mismatch against `esp_spiffs_info()`. Separate PR — it's unrelated to the web UI.

FozzTexx:
> I think what you're looking at is a stale web UI rather than stale code — the firmware on that board is this branch, but `/www` isn't.

Yep, PEBKAC. I always forget to do `uploadfs`

image

cc: @RichStephens

FozzTexx:
I'm good with this PR if @RichStephens is.

FozzTexx (approved):

PR #1624, opened 2026-09-09
Unify device/network for all platforms.

FozzTexx:
Created new NDevice class which 1108s all the common network code into device/NDevice with platforms only creating subclasses for commands that require backward compatibility with legacy implementations.

New Parser class has been created which handles document parsing. JSON & SGML are subclasses of Parser. New parsers can easily be added by creating a new subclass.

PR #1640, opened 2026-09-15
[network] report JSON/SGML parse failures instead of success (#1417)

### LINKED ISSUE #1417: sio_parse_json ignores the parser's return value
### Summary

```cpp
    // lib/device/sio/network.cpp
    void sioNetwork::sio_parse_json()
    {
        transaction_begin(TRANS_STATE::NO_GET);
        json->parse();
        transaction_complete();
    }
```

`FNJSON::parse()` returns `false` when the buffer was empty or
`cJSON_Parse` failed, but the command always completes successfully.
The client learns about the failure only indirectly, when a subsequent
query returns an empty result — indistinguishable from "the queried
path does not exist in valid JSON". A `transaction_error()` on
`parse() == false` would make failures diagnosable.

tschak909:
Fixes #1417.

## The problem

The issue quotes `lib/device/sio/network.cpp`, which no longer exists — #1616 renamed it and #1624 collapsed every per-bus network device into `lib/device/NDevice/`. The bug survived the move and had multiplied into **three stacked layers**, so fixing only the handler would have changed nothing observable:

| Layer | Was |
|---|---|
| `NDevice::fujidev_do_parse` | `_parser->parse();` then unconditional `transaction_success()` |
| `JSONParser::parse` | `_json.parse();` then `RETURN_SUCCESS_AS_FALSE()` |
| `SGMLParser::parse` | `_sgml.parse();` then `RETURN_SUCCESS_AS_FALSE()` |

`fujidev_do_parse` also dereferenced `_parser` with no null check. `_parser` is null before any `NET_OPEN`, after `NET_CLOSE`, and after a failed `NET_OPEN` — so `NET_PARSE` in any of those states **crashed the firmware**. Checked against master with a throwaway host harness: `NET_PARSE` with no channel open SIGSEGVs, so that path was reachable, not theoretical.

## The fix

Result propagates all three layers to `transaction_error()`, plus a sticky `NParser::_parseError` so a subsequent `STATUS` reports `NDEV_STATUS::COULD_NOT_PARSE_JSON` (213). That constant already existed in `status_error_codes.h`; only `GCAL.cpp` set it. Six buses inherit this via `NDevice` (`sio`, `iec`, `adamnet`, `comlynx`, `iwm`, `rs232`), plus DriveWire through `using drivewireNetwork = NDevice;`.

Two design calls worth flagging for review:

- **Parser-side status, not `_protocol->error`.** The GCAL idiom doesn't propagate here: `NetworkProtocolFS::status_file` and `NetworkProtocolTCP::status_client` force `END_OF_FILE` whenever nothing remains, which is always true after a parse drains the body. It would have worked for HTTP and silently failed for everything else.
- **Why 213 matters at all.** `systemBus::transaction_error()` is still a no-op on IEC (`// FIXME - signal error somehow`) and on Mac. Without the STATUS path the fix would be invisible to Commodore users.
- **Guard placement.** Null guards sit *after* `transaction_accept()`: pre-accept yields a NAK (Atari 139, "device not there"), which misdescribes "no channel open". `fujidev_seek` sets that precedent for the same `_parser` null condition. In `fujidev_set_query` the guard sits after the `WILL_GET` payload is drained, or the bus desyncs.

Also included: `NET_QUERY` (`fujidev_set_query`) dropped `fujicore_set_query()`'s `error_is_true` and `transaction_get()`'s result the same way.

## Behaviour changes

- **An empty response body now errors.** `FNJSON::parse()` conflates "empty body" with "malformed JSON", so a 204 or zero-length 200 now fails `NET_PARSE`. This is the most likely source of an "it used to work" report.
- **`NET_PARSE`/`NET_QUERY` without a prior `NET_SET_PARSER` now error.** `fujidev_open` installs a plain `NParser`, whose `parse()`/`setQuery()` already returned error. Pre-#1624 the SIO handler ran the JSON parser regardless of channel mode, so clients that skipped `XIO 252` have been silently no-oping since #1624 — this makes existing breakage diagnosable rather than introducing it. The documented flow (`NET_SET_PARSER` 0xFC, mode in aux2) is unaffected; 0xFC and its values are unchanged from the old `sio_set_channel_mode`.

## Testing

- `ctest --test-dir build` — 5/5 pass
- ESP builds: `fujinet-atari-v1`, `fujiapple-rev0`, `fujinet-adam-v1`
- `coding-standard.py --against master` clean

**rc2014 is out of scope here.** `rc2014Network::rc2014_parse_json()` has the same dropped-result bug, but `lib/device/rc2014/rc2014Network.cpp` does not compile on master — it still references `ProtocolParser`, renamed to `NetworkProtocolFactory` in #1600 (11 syntax errors), and the `fujinet-rc2014spi-rev0` env fails earlier still in `rc2014bus.cpp` on `fujiDeviceID_t` vs `uint8_t`. Nothing there can be compiler-verified, so it is left untouched and will be addressed separately. `H89Network`'s copy is entirely commented out and was likewise left alone.

## Suggested manual check

```basic
OPEN #1,12,0,"N:https://httpbin.org/html"
XIO 252,#1,0,1,"N:"   : REM NET_SET_PARSER, aux2=1 -> JSON
XIO 80,#1,0,0,"N:"    : REM NET_PARSE  -> expect ERROR 144
STATUS #1             : REM expect err byte 213
```

Plus the regression that matters: the same sequence against valid JSON must still parse and query normally.

## Noted, deliberately not fixed here

- `HAVE_LAST_ERROR` is defined nowhere in the tree, so every `lastError = ...` in `NDevice.cpp` is dead code and references a non-existent `protocol->error`/`mode`. Candidate for deletion.
- `fujidev_set_parameter` NAKs pre-accept, inconsistent with the rest of the file.
- rc2014's `NET_PARSE` sends neither ack nor complete when `channelMode != JSON`, leaving the bus hanging.


lib/device/rc2014/rc2014Network.cpp
```diff
@@ -368,7 +368,11 @@ void rc2014Network::rc2014_parse_json()
 
     rc2014_send_ack();
 
-    json.parse();
+    if (!json.parse())
+    {
+        rc2014_send_error();
+        return;
+    }
```
FozzTexx: Don't mess with this file, it's obsolete and needs to be updated to inherit from NDevice, which should be in a separate PR.

tschak909: got it. I'll peel away.

tschak909: removed.

tests/sioNetwork-DSTATS-test.cpp
```diff
@@ -279,3 +282,39 @@ error_is_true SGMLParser::setQuery(const std::string &query) { RETURN_ERROR_AS_T
...
+    pkt.frame.comnd = CMD::NET_SET_PARSER;
+    REQUIRE(dev.processCommand(pkt));
+
+    g_success_count = g_error_count = 0;
+    pkt.frame.comnd = CMD::NET_PARSE;
+
+    REQUIRE(dev.processCommand(pkt));
+    CHECK(g_error_count == 1);
+    CHECK(g_success_count == 0);
+}
```
FozzTexx: I don't know what this is and why it's in here. This test is for Atari only and for validating that the DSTATS command returns an SIO direction for every command that NDevice responds to. This test has nothing to do with parsing.

tschak909: yeah, we can peel that away.

tschak909: removed.

FozzTexx:
I think to go along with this we need a proper test in lib-experimental/testing which is going to exercise the NDevice and make sure that all the error codes from triggering error conditions are correct. Perhaps your intern can open an issue on https://github.com/FozzTexx/fujinet-lib-experimental to remember that the test needs to be created.

FozzTexx (approved):

PR #1642, opened 2026-09-16
Rename FNSGML -> FNHTML, SGMLParser -> HTMLParser

tschak909:
Split out of #1557 at review request: *"If SGML is becoming HTML then the rename should be a PR all on its own."*

Gumbo is an HTML5 tree-construction parser with a CSS selector engine on top; it was never general-purpose SGML, and calling it that has been misleading since it landed. XML gets its own parser in #1557, which is what made the misnomer awkward.

### Purely lexical

| from | to |
|---|---|
| `lib/fnsgml/fnsgml.{h,cpp}` | `lib/fnhtml/fnhtml.{h,cpp}` |
| `FNSGML` | `FNHTML` |
| `lib/device/NDevice/SGMLParser.{h,cpp}` | `HTMLParser.{h,cpp}` |
| `SGMLParser` | `HTMLParser` |
| `PARSER::SGML` | `PARSER::HTML` |
| `SGML_REMAP_CHARS`, `SGML_REMAP_ATASCII_INTERNATIONAL` | `HTML_REMAP_*` |

No behavior change and no wire change: `PARSER::HTML` is still value 2 on `NET_SET_PARSER`. `JSON_DELETE_SGML_TAGS` keeps its name — it means "strip markup tags" and is unrelated to which parser is selected.

### Obsolete devices left alone

`H89Network.h` and `rc2014Network.h` carry an unused `#include "fnsgml.h"`, left in place per review (both are obsolete: they still derive from the pre-1108 `virtualDevice`, and neither `BUILD_H89` nor `BUILD_RC2014` is built by CI). The include dangles after this commit, which costs those two targets nothing they had — dropping the include, or the devices, is a separate question.

### Verified

All six fujinet-pc targets build; ATARI 5/5 tests, APPLE/ADAM 4/4. ESP32 `fujinet-atari-v1` builds, and the binary is **byte-identical to master** (flash 3270867, RAM 122836 on both), which is the strongest available evidence that the rename is lexical.


lib/fnhtml/fnhtml.cpp
```diff
@@ -132,12 +132,12 @@ void FNSGML::resolveQuery()
 /**
  * Character remapping for target platforms (mirrors FNJSON).
  */
-std::string FNSGML::processString(std::string in)
+std::string FNHTML::processString(std::string in)
 {
 #ifdef BUILD_ATARI
```
FozzTexx, line 137: Completely unrelated to this PR, but it needs an issue opened since it's technical debt. Instead of putting `BUILD_` ifdefs in here it should *probably* use the more generic `SYSTEM_BUS.nativeTextToUnicode()` and `SYSTEM_BUS.unicodeTextToNative()` instead. Issue can document suggested change and then someone can explore if it makes sense later.

FozzTexx (approved):

PR #1643, opened 2026-09-16
Mac 68k: boot from floppy and HD20, writable floppy images

dillera:
Follow-up to #1633. Tested on an SE (System 6.0.8) and a 512Ke with the FujiNet as its only disk.

Floppy: sector images (.dsk/.img/DiskCopy 4.2, 400K and 800K) in slot 5 are encoded to GCR at mount time and served like MOOFs; both sizes boot. The Pico behaves like a Sony drive where the ROM cares: READY spin-up and head-settle timing, the disk-changed flag and its clear command, motor off when deselected, spurious ejects from a floating bus ignored, 400K/800K drive identity following the disk.

Floppy writes: for sector images mounted read/write the Pico captures the IWM write stream (new PIO program on WR), the ESP32 decodes it, stores the sector in the image file and patches the encoded track in place. Verified with the Finder writing to a 400K MFS boot disk on the 512Ke. MOOFs stay read-only.

HD20: the ROM's boot-time probe enters the DCD handshake from the device detection state rather than from idle; the Pico now answers it from any state, so a 512Ke boots from the emulated HD20.

Debug: Pico status-line trace (toggle with 't' on its USB console), reflash with picotool without BOOTSEL, host test for the GCR encoder and decoder (tests/mac_gcr_test.cpp).

Docs: docs/mac68k.md, docs/mac68k-floppy-write.md.


tschak909 (changes requested):
@dillera please peel away the .md files. Also intern needs to resolve conflicts due to recent rebase.

PR #1628, opened 2026-09-13
Add COCO_HS_UART support to ESP32UARTChannel; fix DevKitC UART1 RX pin

wdathing:
Adds support for flow control.
Cleaned up some debug messages from my last push.

PR #1650, opened 2026-09-20
[rs232] don't enable UART RTS/CTS hardware flow control implicitly

deltecent:
**Severity:** since #1628, `master` is unusable on RS232 builds. The firmware stops after the boot banner and never reaches `Setup complete`, so SD, config, WiFi and the bus never start. Bisect: `209b0e0e1` (master) bad, `b3dcff67d` (#1628) first bad, `617ab5dc2` good.

**Cause.** #1628 added an unconditional `uart_set_hw_flow_ctrl(UART_HW_FLOWCTRL_CTS_RTS)` in `ESP32UARTChannel::begin()` for any channel with both control pins defined. The RS232 pinmaps (`rs232_rev0.h`, `rs232_s3.h`) define both `PIN_RS232_RTS` and `PIN_RS232_CTS`, so the bus UART picked up CTS/RTS flow control implicitly, TX was gated on CTS, and boot stopped right after the `--~--~--~--` banner.

**Fix.** Remove that call. `uart_param_config()` at the top of `begin()` already applies `conf.uart_config.flow_ctrl`, which defaults to `UART_HW_FLOWCTRL_DISABLE`, so the extra call was redundant and overrode whatever the caller configured. Flow control is now opt-in: a bus that wants it passes `.flowControl(UART_HW_FLOWCTRL_CTS_RTS)` in its `ChannelConfig` (drivewire already has this parked, commented out). No board gets it implicitly, and there are no per-platform `#ifdef`s in `ESP32UARTChannel`.

On these boards RTS/CTS are software-driven GPIO modem-control lines, not UART-peripheral pins, so "both pins are defined" is not evidence that the UART should drive them. `rc2014sio.cpp` (lines 31 and 310) shows the same thing independently: it sets `PIN_RS232_RTS` as an input itself and polls it with `digital_read`. The RS232 board is capable of hardware RTS/CTS, but that also needs the client side to support it, so it must be opted into, never inferred from the pin definitions.

This can't be fixed from the RS232 bus code: #1628's block runs after `uart_param_config()` and keys on the pin numbers, not on `conf.uart_config.flow_ctrl`, so a caller passing `.flowControl(UART_HW_FLOWCTRL_DISABLE)` is overridden, and forcing `rts`/`cts` to `-1` would break the GPIO setup directly above it.

Only `rs232_rev0.h` and `rs232_s3.h` define both pins, so no other board was affected.

**CoCo.** `coco_devkitc.h` (the only `COCO_HS_UART` board) defines neither `PIN_RS232_RTS/CTS` nor `PIN_UART1_RTS/CTS`, so `rts`/`cts` are `-1` there and #1628's block never ran on that board. Removing it does not change CoCo behavior. The CoCo devkitc-centipede build was not tested.

**Not addressed.** `uart_set_pin()` is still called with `UART_PIN_NO_CHANGE` for RTS/CTS, and `begin()` configures those pins as plain GPIO. So setting `.flowControl()` alone does not yet route the UART's RTS/CTS signals; a bus that wants real hardware flow control will need that plumbing added. Left out to keep this fix minimal.

**Tested on hardware:** FujiNet for RS-232 Ver.1b (RS232-S3 build) connected directly to a USB-232 adapter. This configuration worked before the CoCo merge (#1628) and failed to boot after it. With this change it boots again: the FujiBus link comes up, the SD card mounts and lists over the bus, a `.DSK` image mounts and ejects, and WiFi comes up (device answers ping).


lib/hardware/ESP32UARTChannel.cpp
```diff
@@ -81,11 +81,15 @@ void ESP32UARTChannel::begin(const ChannelConfig& conf)
         fnSystem.set_pin_mode(controlPins.cts, gpio_mode_t::GPIO_MODE_OUTPUT);
         fnSystem.digital_write(controlPins.cts, DIGI_LOW);
     }
+#ifdef COCO_HS_UART
+    // Only this board is wired for hardware flow control. Boards that merely route
+    // RTS/CTS as modem-control lines (e.g. RS232) must not have TX gated on CTS.
```
FozzTexx: Do not put per platform `#ifdefs` in this file. If the config is wrong then lib/bus/rs232.cpp needs to pass the correct config.

The FujiNet RS232 device *does* have the RTS/CTS pins wired to GPIO so it is capable of doing RTS/CTS, but it also requires that the client side support it too.

deltecent: Agreed, thanks. Removed the `#ifdef` in 78bff043f.

`uart_param_config()` already applies `conf.uart_config.flow_ctrl` (default `UART_HW_FLOWCTRL_DISABLE`), so the unconditional `uart_set_hw_flow_ctrl()` from #1628 was redundant and overrode the caller's config. I deleted it, so flow control is now opt-in: a bus that wants it passes `.flowControl()` in its `ChannelConfig`, as drivewire's parked lines already do.

On `PINMAP_COCO_DEVKITC`, `rts` and `cts` are -1 (it defines neither `PIN_RS232_*` nor `PIN_UART1_*`), so the block never ran there. It only ran on the RS232 boards, which is what broke boot. Removing it doesn't change CoCo behavior.

Retested on a FujiNet for RS-232 Ver.1b connected directly to a USB-232 adapter: boots again, SD lists over FujiBus, a .DSK mounts, WiFi comes up.

One caveat: `uart_set_pin()` still passes `UART_PIN_NO_CHANGE` for RTS/CTS, so setting `.flowControl()` alone doesn't route those signals yet. That plumbing would be needed before any board uses real hardware flow control.

lib/hardware/ESP32UARTChannel.cpp
```diff
@@ -81,11 +81,6 @@ void ESP32UARTChannel::begin(const ChannelConfig& conf)
         fnSystem.set_pin_mode(controlPins.cts, gpio_mode_t::GPIO_MODE_OUTPUT);
         fnSystem.digital_write(controlPins.cts, DIGI_LOW);
     }
-    if (controlPins.rts >= 0 && controlPins.cts >= 0)
-    {
-        uart_set_hw_flow_ctrl(_uart_num, UART_HW_FLOWCTRL_CTS_RTS, 0);
-        Debug_printv("RTS/CTS flow control enabled");
-    }
```
FozzTexx, line 88: Can't simply delete functionality that was added for a reason. There needs to be a config option to enable it, with the default being disabled.

deltecent: ok, then I will pull this PR and post an issue that the RS232 firmware no longer works on the FujiNet for RS232 Ver.1b device and let someone that understands the firmware better than I do propose a fix.

deltecent:
Closing this PR. Filed https://github.com/FujiNetWIFI/fujinet-firmware/issues/1651 so someone who knows the firmware better can propose a fix.

PR #1652, opened 2026-09-20
Fix infinite loop on EOF and empty lines in fnFTP::read_directory

delysio:
### Summary
Fixes an infinite loop and system freeze when listing FTP directories that end with blank lines or when reaching EOF in `fnFTP::read_directory`.

### How to Reproduce
1. In an Atari environment (or emulator), MyDOS:
   ```text
   NCD 
   ftp://ftp.untergrund.net/
   NDIR
   ```
2. **Observed Behavior (Bug):**
   FujiNet enters an infinite loop consuming 100% CPU in `read_directory`. The Atari receives no SIO response and eventually times out with `Error 138 (Device does not respond) or responses FUJINET NOT RESPONDING`.
3. **Expected Behavior (Fixed):**
   The directory listing completes immediately and displays all entries (`ROBOTS.TXT`, `USERS`, `WELCOME.MSG`, etc.).

### Root Cause
In `lib/network-protocol/FS.cpp`:
```cpp
while (read_dir_entry((char *)entryBuffer.data(), ENTRY_BUFFER_SIZE - 1) == FUJI_ERROR::NONE)
```
The caller loop relies on `read_dir_entry` returning `FUJI_ERROR::NONE` as long as entries exist, and any non-NONE code (specifically `FUJI_ERROR::UNSPECIFIED`) when the directory list is exhausted (EOF).

However, in `fnFTP::read_directory`:
```cpp
getline(dirBuffer, line);
if (line.empty())
    return FUJI_ERROR::NONE; // no more entries
```
When `getline` hits a trailing empty line or EOF, returning `FUJI_ERROR::NONE` caused `FS.cpp` to loop forever, calling `read_directory` on an empty stream.

Furthermore:
- `line = line.substr(0, line.size() - 1);` assumed fixed CRLF endings and could underflow on empty or LF-only lines.
- `return dirBuffer.eof() ? FUJI_ERROR::UNSPECIFIED : FUJI_ERROR::NONE;` at the end caused the last valid entry to be dropped if reading it coincided with `dirBuffer.eof()`.

### Solution
1. Use `while (getline(dirBuffer, line))` to read entries and skip empty lines.
2. Safely strip trailing `\r` with `line.pop_back()`.
3. Return `FUJI_ERROR::NONE` only when a valid entry is successfully parsed by `ftpparse`.
4. Return `FUJI_ERROR::UNSPECIFIED` when no more entries exist, cleanly terminating the directory parser loop.


tschak909:
@delysio we have a new set of coding rules that have been done in the last day or so, to clean up a lot of things like code and PR comment verbosity. They're now a part of AGENTS.md and CONTRIBUTING.md, but the TL:DR here is, 

Trim down the comments for both PR and code comments so we don't drown in a sea of comments :)

RichStephens (approved):
This looks OK to me.

FozzTexx (approved):

PR #1661, opened 2026-09-21
[apple][media] drop the WOZ1 track-loader scratch allocation

tschak909:
`woz1_read_tracks()` allocated a full `WOZ1_TRACK_LEN` buffer (6.6K, `MALLOC_CAP_INTERNAL` on ESP) solely to `fread` six trailing bytes per track into and discard, and never checked it — so a failed allocation meant `fread` into null. A six-byte stack buffer does the job. The real per-track buffers were already SPIRAM and are untouched.

Split out of #1613 per review.

### Testing
`make atari-lwm` 11/11 ctest. `./build.sh -b` for `fujiapple-rev0` — SUCCESS.

**Not flashed, no hardware verification, no test coverage** — nothing parses WOZ in `tests/`. Mounting a real WOZ1 image on an Apple II is what would confirm track loading still works.


RichStephens (approved):
Makes sense to me.

PR #1666, opened 2026-09-21
[http] check the url-decode buffer allocation in the config POST handler

tschak909:
`process_config_post()` malloc'd `postlen + 1` and passed the result straight to `url_decode()` with no null check — a write through null on a failed allocation. `postlen` comes from the request body, so it's reachable over the network by anyone who can load the web UI. Returns `-1` now.

Caveat: `mgHttpService.cpp` tests for a negative return, but the ESP caller in `httpService.cpp` assigns `ret` and falls through to its redirect without checking. So on ESP the POST is silently ignored rather than applied — better than the write, but that caller needs its own fix, deliberately not in this PR.

Split out of #1612 per review.

### Testing
`make atari-lwm` 11/11 ctest. `./build.sh -b` for `fujinet-atari-v1` — SUCCESS.

**Not flashed, no hardware verification, no test coverage.** I didn't induce the allocation failure; that needs the device already near exhaustion.


nwah (approved):
things are that tight that we hit this eh?

PR #1667, opened 2026-09-21
[rs232] the bus owns the reply timeout, not the media type

tschak909:
`IOChannel`'s read timeout is a short per-byte window sized for frame recovery. That's wrong for a command/reply exchange whose peer can stall — an RP2040 cart commits a 4K flash sector (~400ms on W25Q) before ACKing the chunk that filled it, so the reply lands long after the window gave up.

`_transaction()` now wraps the write/read pair in a `ScopedReplyTimeout` (2000ms, restored by destructor). `service()` keeps its short window.

Split out of #1605 because it widens the window for every RS232 device; #1605 depends on it. A peer that used to fail fast now takes up to 2s to do so.

### Testing
`make atari-lwm` 11/11 ctest. `fujinet-rs232-rev0` builds.

**Not flashed, no hardware verification, no test coverage.** The 400ms stall was seen in emulation, not on silicon.


lib/bus/rs232/rs232.h
```diff
@@ -135,6 +141,17 @@ class systemBus : public SystemBusBase
...
+    {
+        IOChannel *ch;
+        double saved;
+        explicit ScopedReplyTimeout(IOChannel *c) : ch(c), saved(c->readTimeout())
+        {
+            ch->setReadTimeout(RS232_REPLY_TIMEOUT_MS);
+        }
+        ~ScopedReplyTimeout() { ch->setReadTimeout(saved); }
+    };
+
```
FozzTexx: I can't make sense of this obfuscation.

tschak909: Dropped the RAII wrapper — it is now a plain save/set/restore around the reply read, inline in `sendCommand()`.

FozzTexx:
> IOChannel's read timeout is a short per-byte window sized for frame recovery. That's wrong for a command/reply exchange whose peer can stall — an RP2040 cart commits a 4K flash sector (~400ms on W25Q) before ACKing the chunk that filled it, so the reply lands long after the window gave up.

Now that I can see what's going on, it seems like the correct fix is not to send such a huge chunk. Increasing timeouts is just a bigger hammer solution. The fujiversal firmware already allows sending a ROM over as multiple commands and that's how it's done in disk.cpp. I'm not sure where the notion that 4k is being sent all at once came from since `push_stream()` is sending in 512 byte chunks.

tschak909:
You're right, and my PR body was wrong twice over.

`push_stream()` does send 512-byte chunks — `uint8_t buffer[DISK_SECTORBUF_SIZE]`, and that's 512 in `lib/media/rs232/diskType.h`. The 4K I wrote is the cart's flash **sector**, not the wire chunk. Bad wording on my part, and it's what sent this sideways.

The stall is in the astrocade cart firmware over in #1605: the 512-byte chunks accumulate into a 4K sector buffer, and every 8th chunk fills it and triggers `flash_range_erase` + `flash_range_program` *before* the ACK goes out. Smaller chunks wouldn't help — 4K is the part's erase granularity, not something the wire chooses, so 256-byte chunks would just move the stall from every 8th chunk to every 16th.

Second error in that body: "it widens the window for every RS232 device." It doesn't. `sendCommand()` on the RS232 bus has exactly one consumer, `push_stream()` itself; `service()` and the device command path call `readBusPacket()` directly and never see it.

But that cuts against this PR rather than for it. The only peer that stalls in that path is the astrocade cart, which isn't merged. Intellivision — the one shipped RS232 cart — buffers the ROM into cart RAM in `bootmap_rom_data()` and ACKs immediately; its flash use is the mass-storage drive, not the command path. So there's no peer on master this helps.

Closing. The fix belongs on the cart: pre-erase the store so the in-path cost is programming only, ~11ms per 4K, comfortably inside the existing 100ms window. If a residual wide window is still needed for a second push against a dirty store, it goes in #1605 at that one call site where the context is visible, not as a bus-wide constant here.

PR #1668, opened 2026-09-21
[ftp] Pin the FTP LIST entry loop with host tests

tschak909:
Tests for #1652, which nothing covered. #1652 has merged; this is rebased on master.

**Two commits, and the first one is a pure move — read it that way.** GitHub renders it as a 600-line rewrite of `fnFTP.cpp`, which it isn't:

1. `split ftpparse and the LIST loop out of fnFTP.cpp` — no behaviour change. D. J. Bernstein's `ftpparse` was sitting inline at the top of `fnFTP.cpp`; it moves to `lib/ftp/ftpparse.{h,cpp}` byte-for-byte, the only additions being the header guard, the declaration, the include and the comments. `read_directory`'s body moves to `lib/ftp/ftp_dir_parse.{h,cpp}` verbatim, taking `dirBuffer` as an `istream`. `fnFTP::read_directory` becomes a one-line delegate. It had to move: `fnFTP.h` pulls in `Protocol.h` -> `bus.h`, and `ftpparse` was file-local, so the loop couldn't link on its own.
2. `pin the LIST entry loop with host tests` — `tests/FtpDirectoryTests.cpp`, 6 cases, 50 assertions, wired in as the `ftp_directory_tests` ctest. Covers termination on blank lines/EOF, the dropped last entry, LF-only names, skipping lines ftpparse rejects, and the `ftp.untergrund.net` listing from the report.

`./build.sh -p ATARI` green at both commits — 11/11 at the move, 12/12 at the tip. Verified the tests can fail: built against the pre-#1652 loop, every case fails and the blank-line case reproduces the hang. No ESP32 board build, nothing flashed.

Client-side test: FozzTexx/fujinet-lib-experimental#118.


RichStephens (dismissed):
Tested with latest fujinet-lib-experimental. Intended tests pass there, and all the firmware build tests that were added here pass as well. No reason not to merge this now that #1652 has merged.

FozzTexx:
> Tested with latest fujinet-lib-experimental. Intended tests pass there, and all the firmware build tests that were added here pass as well. No reason not to merge this now that #1652 has merged.

Needs to be rebased onto master.

FozzTexx:
> this is rebased on master and is now just the tests

This seems to be completely refactoring lib/ftp, and is not "just the ~~facts~~ tests, ma'am."

RichStephens:
> > this is rebased on master and is now just the tests
> 
> 
> 
> This seems to be completely refactoring lib/ftp, and is not "just the ~~facts~~ tests, ma'am."

Maybe intern lied, then. It said the only difference between this and #1652 was only the tests.

FozzTexx:
I'm not a fan of refactoring everything just to add a test.

tschak909:
we CAN drop this side of it, because we have a client side test.

FozzTexx:
> we CAN drop this side of it, because we have a client side test.

Close it.

PR #1669, opened 2026-09-21
[ci] stop building every commit twice, and cancel superseded runs

tschak909:
`Build FujiNet-PC` is a 48-job matrix, 18 of them macOS, where GitHub caps concurrency at 5. Two things kept that queue full of dead work:

- `push` and `pull_request` both fired on `"**"`, so every commit on an open PR built the full matrix twice on the same SHA.
- No `concurrency:` group, so superseded runs held their queue slot.

This morning: 19 queued runs, 205 macOS jobs waiting, oldest ~10 hours. Nothing had failed — 555 completed, zero failures. About 8 of the 19 runs were redundant.

Fix, applied to the three workflows that build on PRs (`build-fujinet-pc.yml`, `autobuild.yml`, `pico-carts.yml`): restrict `push` to `master`, and add

```yaml
concurrency:
  group: ${{ github.workflow }}-${{ github.event_name }}-${{ github.event.pull_request.number || github.ref }}
  cancel-in-progress: ${{ github.event_name == 'push' || github.event_name == 'pull_request' }}
```

`github.event_name` is in the key on purpose: the nightly `schedule` run publishes the release and its `github.ref` is `refs/heads/master`, so a simpler ref-only key would let a master push cancel an in-flight nightly. Master pushes publish nothing themselves, so collapsing a burst of merges to the tip is safe.

Tradeoff: a branch with no PR open no longer builds on push. `autobuild.yml` gains `workflow_dispatch` (the other two already had it).

Not included: trimming the macOS leg of the matrix (3 variants × 6 targets = 18 of the 48 jobs). That's a coverage call, separate PR.


RichStephens (approved):
Looks good to me.

.github/workflows/autobuild.yml
```diff
@@ -1,18 +1,25 @@
...
 on:
+  # Only master builds on push; branch pushes are covered by pull_request, which
+  # used to run the identical SHA a second time. workflow_dispatch covers a
+  # branch that has no PR open yet.
   push:
     # Do not build tags. Do not build release branch
     branches:
-      - "**"
-      - "!release"
+      - master
```
FozzTexx, line 10: No, this will prevent checks from running whenever people push to another branch and want to run checks before opening a PR.

FozzTexx (changes requested):
"Checks running twice" is only an issue for people that are creating branches directly on FujiNetWiFi, not on their own private fork.

Having checks run when people push to a different branch on their own fork is **essential** so that they can check that everything is good *before* opening a PR.

tschak909:
so do we want to just close this?

FozzTexx:
> so do we want to just close this?

If its only purpose was to make it so that checks won't get run on a push, then yes it should be closed.

PR #1672, opened 2026-09-21
[fujiversal] flash the ESP32 and its cartridge MCU from one image

tschak909:
# Summary

Fujiversal boards are two chips — an ESP32-S3 and an RP2040/RP2350/RP2354 cartridge joined by USB — but only the ESP32's UART is exposed, so the cartridge had to be flashed separately over its own BOOTSEL port. This embeds the cartridge firmware in `firmware.bin` and has the ESP32 push it over PICOBOOT at boot, so one zip and one FujiNet-Flasher run cover both chips.

## Changes

- Cartridge firmware is built alongside the ESP32 build and embedded as rodata. No partition-table, zip-layout or flasher-protocol change: `release.json` gains an optional `companion` array describing what is embedded, and nothing else moves.
- New `lib/hardware/`: `PicobootClient` (the BOOTSEL wire protocol, ported from picotool onto ESP-IDF `usb_host`), `fnPicoUpdater` (policy, NVS bookkeeping, BOOTSEL recovery), `fnUsbHost` (shared `usb_host_install`). Called from `main_setup()` before the bus, all behind `CONFIG_USB_PICOBOOT_HOST_ENABLED`.
- `ACMChannel` no longer installs the USB host itself. When it finds the host already running it recycles the root port, because `NEW_DEV` is only sent at enumeration and never replayed — without it `begin()` waits forever for a cartridge that enumerated while the updater had the port, which is the normal case right after a reflash.
- The cartridge is only rewritten when the embedded image's sha256 differs from what NVS records was last flashed. Flashing erases NVS, so a flasher run reflashes exactly once and later boots take the fast path.
- `build_pico.py` gains `pico_chip`, `pico_flash_base`/`pico_flash_limit`, `pico_prebuild`, a submodule check, and records sha256/chip/flash bounds in the blob table. It fails the build if a board embeds an image without `CONFIG_USB_PICOBOOT_HOST_ENABLED`, or if an image would be written past its limit.
- New boards `fujiversal-drivewire` (CoCo bus) and `fujiversal-msx`, both building from `pico/fujiversal`, added as a pinned submodule. The MSX CONFIG ROM is checked in prebuilt (building it needs z88dk, which CI has not got).
- Intellivision cartridge: enables the 1200-baud BOOTSEL convention, and drops the activity-LED argument from the mailbox doorbell so an RP2350 does not take the RISC-V boot path.
- CI builds the three Fujiversal boards, installing the ARM toolchain only for those jobs.

## Notes

The first two commits are @nwah's `add-rp-and-esp32s3-build` (#1532), which was merged into its own branch and never reached `master` — #1531 was closed. They are the foundation this builds on, so they ride along here rather than being rewritten.

The 1200-baud path is worth a look: the SDK compiles it out by default for a project that links `tinyusb_device` directly, as the Intellivision cartridge does, so linking `pico_usb_reset` alone silently does nothing. `arm-none-eabi-nm <elf> | grep line_coding_cb` is the only way to be sure it is really there.

Happy to split this into a series (groundwork / USB host refactor / updater / boards / CI) if you would rather review it that way.

## Testing

- Firmware build (`./build.sh -b`): `fujiversal-intv`, `fujiversal-drivewire`, `fujiversal-msx`, plus `fujinet-atari-v1`, `fujinet-rs232-s3`, `fujiversal-rs232`, `fujiversal-o2` as regression.
- `FUJINET_SKIP_PICO=1 ./build.sh -b` on `fujiversal-intv`: links the stub, flash drops by the image size.
- Cartridge firmware: all three build; `Minty_fujicard.elf` checked with `nm` for `tud_cdc_line_coding_cb`.
- Release zips (`./build.sh -z`): `fujiversal-msx` and `fujinet-rs232-s3`.
- Flasher (companion notice + `PICOFW:` highlighting): FujiNetWIFI/fujinet-flasher#16, tested against both zips above.

**Expected:** a Fujiversal board flashed with the FujiNet-Flasher logs `PICOFW:` lines on the same UART, ending `OK ... flashed and verified` then `companion is up`, and on the next boot `up to date ... no reflash needed`. Every non-Fujiversal board builds unchanged and its `release.json` has no `companion` key.

**Not tested:** none of it has run on hardware. Specifically unverified: `PC_REBOOT2` on a real RP2350, whether `PC_READ` after `PC_EXIT_XIP` returns live flash for the verify pass, whether the host controller reliably notices the cartridge's soft reset after the 1200-baud request, and the GPIO BOOTSEL forcing on the Intellivision board. No test coverage — none of this is reachable from the PC build.


tschak909:
Split into a series, one concern each, every branch off master:

- #1673 build a companion MCU image into the firmware (1/6)
- #1674 share the USB host install between its users (2/6)
- #1675 flash the cartridge MCU from the ESP32 at boot (3/6)
- #1676 let the host reboot the cartridge into BOOTSEL (4/6)
- #1677 add the CoCo-bus and MSX boards (5/6)
- #1678 build the boards in CI, and document the flashing path (6/6)

#1674 and #1676 stand alone and can land whenever. The rest land in order: 1, 3, 5, 6.

Same content — the six branches together reproduce this branch's tree exactly, file for file.

PR #1673, opened 2026-09-21
[fujiversal] build a companion MCU image into the firmware (1/6)

tschak909:
# Summary

A Fujiversal board is two chips: an ESP32-S3 and an RP2040/RP2350/RP2354 cartridge joined by USB, with only the ESP32's UART exposed. This is the build-system half of flashing both from one image — it builds the cartridge firmware alongside the ESP32's and carries it inside `firmware.bin`, recording what is needed to flash it later. Nothing consumes it yet.

First of six; the rest are listed at the bottom.

## Changes

- Carries @nwah's `build_pico.py`/`build_merge.py` work (#1532), which was merged into its own branch and never reached master — #1531 was closed. It is the foundation for the rest.
- `pico/fujiversal` (MSX and CoCo cartridge firmware) as a pinned submodule; the MSX CONFIG ROM prebuilt and checked in beside it, since building it needs a Z80 toolchain CI has not got. `pico/tools/rom2h.py` renders a ROM to the C header that tree includes — `xxd` without `xxd`, which is not installed everywhere.
- PICOBOOT protocol headers, byte-identical from pico-sdk 2.3.0, so the ESP32 build needs no pico-sdk.
- `build_pico.py` records an image's sha256, chip and permitted flash range. New `pico_chip` (no default — the chip families reboot with different commands and enumerate under different USB IDs, so a guess fails on hardware), `pico_flash_base`/`pico_flash_limit` (the Intellivision cartridge keeps a 1 MB LittleFS of user ROMs and saves above its image; overrunning it is data loss, so it fails the build), and `pico_prebuild`.
- `release.json` gains an optional `companion` array. Not entries in `files` — these live inside `firmware.bin` and have no flash offset. Boards without a companion are byte-for-byte unaffected.

## Notes

`CONFIG_USB_PICOBOOT_HOST_ENABLED` appears on `fujiversal-intv` here with no consumer, and `build_pico.py` enforces the pairing. Deliberate: without it a board would carry an image and have no code to flash it, which nothing else would catch. The consumer arrives in #3 of the series.

Switching between a branch with `build_pico.py` and one without leaves the gitignored `lib/hardware/fn_pico_blob_data.cpp` behind, which then fails to compile. `rm` it, or `./build.sh -c`.

## Testing

- `./build.sh -b`: `fujiversal-intv` (embeds the 115K cartridge image), `fujinet-rs232-s3` as regression.
- `./build.sh -z`: `fujiversal-msx` and `fujinet-rs232-s3`; confirmed the latter's `release.json` gains no keys and lists the same five files at the same offsets.
- `build_pico.py --print-config` / `--dry-run`, the flash-limit and missing-define guards, and the uninitialised-submodule message.

**Expected:** every non-Fujiversal board builds and packages exactly as before. `fujiversal-intv` gains ~115K of flash and nothing else changes at runtime.

**Not tested:** not run on hardware — nothing here has runtime behaviour to run. No test coverage; none of this is reachable from the PC build.

## The rest of the series

Each is a separate branch off master, landing in this order:

1. this one — build system
2. share the USB host install (independent of this one; can land in either order)
3. the updater that flashes the cartridge — needs 1 and 2
4. cartridge 1200-baud BOOTSEL (independent, Intellivision cartridge firmware only)
5. CoCo and MSX boards — needs 1 and 3
6. CI and docs — needs 5


tschak909:
Evidence, RP2350 firmware is in the firmware ZIP ready for use. 

$ unzip -p firmware/fujinet-FUJIVERSAL-INTV-1.6.2-dev.zip release.json | jq .companion

```json
[
  {
    "name": "intv_fw",
    "chip": "rp2350",
    "size": 115276,
    "sha256": "d965ce7e4b5e35663bb709f4fac20bd697f377c467e3b7da7a25c6e516199446",
    "flash_base": "0x10000000",
    "flash_limit": "0x10100000",
    "source": "pico/intellivision/firmware",
    "source_rev": "in-tree"
  }
]
```
[fujinet-FUJIVERSAL-INTV-v1.6.2-dev.zip](https://github.com/user-attachments/files/32484853/fujinet-FUJIVERSAL-INTV-v1.6.2-dev.zip)

tschak909:

PR #1675, opened 2026-09-21
[fujiversal] flash the cartridge MCU from the ESP32 at boot (3/6)

tschak909:
# Summary

Pushes the cartridge firmware embedded in the ESP32 image across the USB link to the cartridge, so one firmware zip and one FujiNet-Flasher run cover both chips.

Third of six. **Needs #1673 and #1674 first** — it will not compile without them, so CI here is red until they land.

## Changes

- `PicobootClient`: the protocol an RP-series bootrom offers in BOOTSEL mode, ported from picotool's `picoboot_connection.c` onto ESP-IDF's `usb_host`. Protocol only — `PicoUpdater` opens and claims the device and hands it over, so one USB client serves both discovery and flashing.
- `PicoUpdater`: runs from `main_setup()` before the bus, because the bus blocks waiting for the cartridge's CDC endpoint, which a cartridge in BOOTSEL never presents — a blank one would hang there forever. Flashes anything found in BOOTSEL outright; otherwise reflashes only when the embedded image's sha256 differs from what NVS records was last flashed. Flashing a FujiNet erases NVS, so a flasher run reflashes exactly once and later boots leave the cartridge alone.
- All of it behind `CONFIG_USB_PICOBOOT_HOST_ENABLED`, which only the Fujiversal boards set.

## Notes

Four things differ from the earlier prototype on `intv-combined-flash`, and are the parts worth reviewing:

- RP2350/RP2354 as well as RP2040 — different BOOTSEL product IDs, and `PC_REBOOT` is RP2040-only, so RP2350 needs `PC_REBOOT2` with a non-zero delay. The FujiCard is RP2354.
- Every sector is read back and compared. A bad write otherwise looks exactly like a good one until the cartridge fails to boot, by which point nothing is left that can say why.
- A timed-out transfer is reclaimed with halt/flush/clear rather than abandoned — it is still owned by the host stack, and the reboot acknowledgement times out on every successful flash by design, so abandoning it would be the normal outcome of the happy path.
- A failed command asks the bootrom for its own status code, so a failure reads as `BAD_ALIGNMENT` rather than "it stalled".

A failed flash leaves the cartridge in BOOTSEL and carries on booting, which the next boot recovers from unaided.

## Testing

- `./build.sh -b` on `fujiversal-intv`, on top of #1673 and #1674. RAM 20.2%, flash 31.3%.
- `FUJINET_SKIP_PICO=1`: links the stub, flash drops by the image size, the updater still compiles and logs `SKIP`.

**Expected:** on a Fujiversal board, `PICOFW:` lines on the monitor ending `OK ... flashed and verified` then `companion is up`; on the next boot, `up to date ... no reflash needed`. On every other board, nothing compiles in and nothing changes.

**Not tested:** none of it has run on hardware, which is the whole of this PR's behaviour. Specifically unverified: `PC_REBOOT2` on a real RP2350, whether `PC_READ` after `PC_EXIT_XIP` returns live flash for the verify pass, whether the host controller reliably notices the cartridge's soft reset after the 1200-baud request, and the GPIO BOOTSEL forcing. No test coverage — none of this is reachable from the PC build.


tschak909:

PR #1676, opened 2026-09-21
[intv][pico] let the host reboot the cartridge into BOOTSEL (4/6)

tschak909:
# Summary

Setting the USB serial line to 1200 baud is the convention picotool and every other pico tool uses to mean "reboot into the bootloader". This cartridge did not implement it, so the only ways in were the mailbox doorbell — which needs working firmware and an Intellivision to poke it — and the hardware RUN/BOOTSEL lines.

Fourth of six, but independent: it stands on its own and can land in any order. Useful by itself, since it also lets picotool reflash the cartridge from a PC.

## Changes

- Link `pico_usb_reset` and define `PICO_ENABLE_USB_RESET_VIA_BAUD_RATE=1`.
- The mailbox doorbell no longer asks for an activity LED. On RP2350 the SDK works around an erratum by rebooting into RISC-V USB boot whenever one is named, and this cartridge is an RP2354. PICOBOOT works either way, but there is no reason to take a different boot path to blink.

## Notes

The explicit `=1` is the point of the change, not boilerplate: `pico_usb_reset` defaults it **off** for a project that links `tinyusb_device` directly, as this one does, so linking the library alone compiles the callback out and nothing tells you. `arm-none-eabi-nm Minty_fujicard.elf | grep line_coding_cb` is the only way to be sure.

Only the baud-rate path is enabled. The vendor reset interface would need descriptor changes and a new interface, and buys nothing here.

## Testing

- `cmake -DPICO_BOARD=fujicard -DCMAKE_BUILD_TYPE=Release -G Ninja && ninja`, against pico-sdk 2.3.0.
- `nm` on the resulting ELF finds `tud_cdc_line_coding_cb`, and its disassembly compares against 1200 and calls `rom_reset_usb_boot_extra(-1, 0, false)`. Absent before this change.

**Expected:** `picotool reboot -f -u` (or any 1200-baud touch) drops the cartridge into BOOTSEL. Nothing else changes: same USB descriptors, same product ID, same mailbox behaviour.

**Not tested:** not run on hardware — verified in the built ELF only. No Intellivision was attached, so the doorbell path was not exercised end to end. No test coverage for this tree beyond the `pico-carts.yml` build.


tschak909:

PR #1678, opened 2026-09-21
[fujiversal] build the boards in CI, and document the flashing path (6/6)

tschak909:
# Summary

Builds the three Fujiversal boards in CI, and documents how a board with two chips is flashed.

Last of six. **Needs #1677 first** (it builds the boards that PR adds).

## Changes

- `FUJIVERSAL-INTV`, `-DRIVEWIRE` and `-MSX` added to the autobuild, nightly and release matrices, each with its own release ini overlay. Those jobs need an ARM cross-compiler and the pico-sdk; a matrix `include` marks them and the two extra steps are gated on it, so every other board's job is unchanged and pays nothing. The SDK is pinned to 2.3.0, as `pico-carts.yml` already does.
- Every source checkout now fetches submodules, since the MSX and CoCo cartridge firmware lives in one.
- `pico-carts.yml` builds all three cartridge firmwares through `build_pico.py` — where a broken ini key, a missing ROM or an uninitialised submodule should surface, at a fraction of the cost of a full ESP-IDF build.
- `docs/fujiversal-flashing.md`: the boot sequence, the `PICOFW:` log lines, when a reflash happens and why a hand-flashed development cartridge is left alone, the four ways into BOOTSEL, and the four places to touch when adding another board.

## Testing

- Workflow YAML parsed and the matrices and gated steps inspected; the overlays follow the existing `platformio.release-*.ini` pattern.

**Expected:** three new jobs per workflow; existing jobs unchanged apart from fetching submodules.

**Not tested:** the workflows have not run — nothing here can be exercised until it is on a branch GitHub builds, and the new jobs will only go green once the rest of the series has landed. Nothing else in this PR is executable.


tschak909:

PR #1682, opened 2026-09-21
[fujiversal] share the USB host install between its users (2/6)

tschak909:
# Summary

`usb_host_install()` may be called exactly once — a second call asserts — but a Fujiversal board has two things that want the host up: `ACMChannel`, which carries bus traffic to the cartridge, and the updater that flashes that cartridge before the bus starts. Moves the install into `fnUsbHost`, where whichever runs first does it.

Second of six, but independent of the others: it can land in any order relative to #1681.

## Changes

- `fnUsbHost` owns `usb_host_install()` and the event-pump task, which keeps the name `usb_lib` (`setServicePriority()` finds it by name) and is still created at the caller's priority, so the boot-priority boost is unchanged.
- `usbHostRecycleRootPort()`, which `ACMChannel` calls when it finds the host already running. `NEW_DEV` is only sent at the moment a device enumerates and is never replayed to a client that registers later, and the CDC-ACM driver raises its new-device callback from that event alone — so a cartridge that enumerated while something else had the port would have announced itself to nobody, and `begin()` would wait forever for a device sitting right there.

## Notes

On its own this changes nothing observable: boards with neither USB-host feature compile `fnUsbHost` to nothing, and where only `ACMChannel` exists nothing installs the host first, so the recycle never runs. It only starts mattering with the updater (#3 of the series).

## Testing

- `./build.sh -b`: `fujiversal-rs232` (USB-host board, exercises the moved install), `fujinet-rs232-s3` and `fujinet-atari-v1` as regression.

**Expected:** identical behaviour on every board today — same task name, same priority, same install arguments.

**Not tested:** not run on hardware. The recycle path is unreachable until something else installs the host first, so nothing here exercised it. No test coverage; `lib/hardware` USB code is not in the PC build.


lib/hardware/fnUsbHost.h
```diff
@@ -0,0 +1,59 @@
...
+// registering after a device has already enumerated will not be told about
+// it (see usbHostRecycleRootPort()).
+//
+// Call before registering a usb_host client, and as early as possible: a
+// device attached at power-on can finish enumerating before a late client is
+// listening.
+//
+// The event task is named "usb_lib"; ACMChannel::setServicePriority() finds
+// it by that name, so do not rename it.
+bool usbHostEnsureInstalled(UBaseType_t event_task_priority);
```
nwah, line 40: nitpick: why are these randomly camelCase?

lib/hardware/fnUsbHost.h
```diff
@@ -0,0 +1,59 @@
...
+// moment a device enumerates (usb_host.c's USBH_EVENT_NEW_DEV handler), and
+// is never replayed to a client that registers afterwards. So a client that
+// comes up second -- the CDC-ACM driver on a board where PicoUpdater already
+// brought the host up and enumerated the companion -- would otherwise wait
+// forever for a device that is sitting right there, already enumerated.
+// Recycling the port makes it announce itself again.
+//
+// Costs roughly half a second and is only correct while no one holds an open
+// handle to a device on the port.
+void usbHostRecycleRootPort();
```
nwah, line 55: ditto camelCase

nwah (approved):

PR #1685, opened 2026-09-21
[fujiversal] add the CoCo-bus and MSX boards (5/6)

tschak909:
# Summary

Adds the two remaining Fujiversal designs, `fujiversal-drivewire` (CoCo bus) and a new `fujiversal-msx`, both building their cartridge firmware from the `pico/fujiversal` submodule for the RP2350B prototype boards.

Fifth of six. **Needs #1681 and #1683 first.**

## Changes

- Board inis for `coco_proto_260402` and `msx_proto_260402`, each rendering its cartridge ROM to the C header that tree includes. The build directory has to be `build/<BOARD>` because that is where its `board_defs.h` looks for the generated header.
- A pinmap for the MSX board, matching the existing Fujiversal siblings.
- `fujiversal-intv`, `-drivewire` and `-msx` all set `platform_name`. They share `build_platform` with other boards, which the release zip is named after, so without it several boards produce a zip of the same name. `-drivewire` also gains `build_board`, which `build_pico.py`'s guard needs.

## Testing

- `./build.sh -b`: `fujiversal-drivewire` and `fujiversal-msx`, on top of #1681 and #1683.
- Cartridge firmware for both boards via `build_pico.py`; 83K and 107K.
- `./build.sh -z` on `fujiversal-msx`: zip named `fujinet-FUJIVERSAL-MSX-*`, same five files at the same offsets, `release.json` naming the embedded image and the submodule commit it was built from.

**Expected:** two new boards appear in `./build.sh -S` and build. No existing board changes.

**Not tested:** neither board has been flashed — no CoCo or MSX hardware attached, and the cartridge PCBs are prototypes. The RP2350B modules declare 16 MB of flash from their board definition; a smaller physical part would not be caught by `pico_flash_limit`. No test coverage.


nwah (commented):
Do we need these/equivalent that the INTV has?
```c
#define PIN_RP2040_RUN           GPIO_NUM_4
#define PIN_RP2040_BOOTSEL       GPIO_NUM_5
```

tschak909:
Yes.

RichStephens (approved):

PR #1686, opened 2026-09-21
[fujiversal] build the boards in CI, and document the flashing path (6/6)

tschak909:
# Summary

Builds the three Fujiversal boards in CI, and documents how a board with two chips is flashed.

Last of six. **Needs #1681 and #1685 first**: #1681 provides `build_pico.py`, which the new `pico-carts.yml` step invokes, and #1685 provides the MSX board ini that the new matrix entry builds. Until both land, this PR's CI is red for those two reasons and nothing else.

## Changes

- `FUJIVERSAL-INTV`, `-DRIVEWIRE` and `-MSX` added to the autobuild, nightly and release matrices, each with its own release ini overlay. Those jobs need an ARM cross-compiler and the pico-sdk; a matrix `include` marks them and the two extra steps are gated on it, so every other board's job is unchanged and pays nothing. The SDK is pinned to 2.3.0, as `pico-carts.yml` already does.
- Every source checkout now fetches submodules, since the MSX and CoCo cartridge firmware lives in one.
- `pico-carts.yml` builds all three cartridge firmwares through `build_pico.py` — where a broken ini key, a missing ROM or an uninitialised submodule should surface, at a fraction of the cost of a full ESP-IDF build.
- `docs/fujiversal-flashing.md`: the boot sequence, the `PICOFW:` log lines, when a reflash happens and why a hand-flashed development cartridge is left alone, the four ways into BOOTSEL, and the four places to touch when adding another board.

## Testing

- Workflow YAML parsed and the matrices and gated steps inspected; the overlays follow the existing `platformio.release-*.ini` pattern.

**Expected:** three new jobs per workflow; existing jobs unchanged apart from fetching submodules.

**Not tested:** the new jobs cannot pass in isolation and have not. Observed on this branch: `pico cartridges` fails because `build_pico.py` does not exist yet (#1681), and `FujiNet CI` fails on `FUJIVERSAL-MSX` because that board ini does not exist yet (#1685). Both clear once those land. Nothing else here is executable.


FozzTexx:
Shouldn't this PR be the first before any of the others? The very first thing that needs to happen in order to flash the pico is to be able to build the pico firmware. None of the other steps for flashing the pico from the ESP32 matter if there's no way to build the pico firmware.

tschak909:
@FozzTexx the boards on 1685 need to go in first, as this references them.

Of the three draft PRs at the moment:

1683 is referenced by 1681 and 1682 (which are now in.)
1685 is referenced by 1681, and 1683 (of which 1681 is now in)
1686 is referenced by 1685.

RichStephens:
I was able to build and flash combined firmware successfully (fujiversal-drivewire, coco3 prototype board).

It took a little doing to install the pico dev kit and ninja, but the build and flash worked.

RichStephens (approved):

RichStephens:
@tschak909 Looks like there's some cleanup needed here.  Builds are failing saying the msx platformio ini file doesn't exist?

PR #1687, opened 2026-09-22
Fix Atari and CoCo file copy

RichStephens:
Tested with disk tests on coco and atari.  Passed on both.

image

image

I thought at first this would fix 1655.  But no, it doesn't. There's an off-by-one error in host slot numbering somewhere that breaks file copy in atari config.  So, the reported crash didn't happen in my test, but it didn't copy anything either, because it was pointing to the wrong hosts.

This PR works fine, though.


lib/device/sio/sioFuji.cpp
```diff
@@ -197,7 +197,7 @@ void sioFuji::sio_set_baudrate(const FujiSIOPacket &packet)
 }
 
 // Do SIO copy
-void sioFuji::sio_copy_file(const FujiSIOPacket &packet)
+void sioFuji::fujidev_copy_file(const FUJI_COMMAND_PACKET &packet)
```
FozzTexx, line 200: Not possible to toss the guts of it and make it call the fujiDevice::fujicmd version like CoCo does?

RichStephens: Yes.  Good catch.

RichStephens: Done and tested.

FozzTexx (approved):

PR #1688, opened 2026-09-22
Update Atari config disk to fix file copy

### LINKED ISSUE #1655: [Bug]: File copy (CMD::FUJI_COPY_FILE $D8) crashes FujiNet in official Config program (Access Violation 0xC0000005)
### Describe the bug
When attempting to copy a file using the copy feature (`C`) in the official FujiNet Configuration Program (`autorun.atr`) on Atari SIO, FujiNet immediately crashes with an Access Violation (exit code `3221225477` / `0xC0000005` in `ucrtbase.dll` on PC, panic/crash on ESP32), causing the Atari to hang.

The crash occurs the moment the copy operation is confirmed, before any data transfer starts.

### Environment
- **Hardware/Emulator:** Atari 8-bit / Altirra with FujiNet-PC (nightly build).
- **Software:** Official FujiNet Config disk (`autorun.atr`), standard Atari OS (no third-party DOS or extensions).
- **Operation:** Copying `anode/anode.xex` from `tnfs://fujinet.pl` to the local SD card.

### Steps to Reproduce
1. Boot the Atari with the official FujiNet Config disk (`autorun.atr`).
2. Have `fujinet.pl` configured as a host slot and `SD` configured as an SD card host slot.
3. Open `fujinet.pl`, navigate to the `anode` directory.
4. Move the cursor over `anode.xex` and press `C` to initiate copy.
5. In the target slot selection screen, select the `SD` slot and press `RETURN` to confirm the copy.
6. Observe that FujiNet immediately crashes.

### FujiNet Log Excerpt
```text
21:12:50.041242 > CF: 70 f5 00 00 66
21:12:50.041750 > sioFuji::fujicmd_process() called, baud: 19200
21:12:50.042031 > ACK!
21:12:50.042036 > Fuji cmd: CLOSE DIRECTORY
21:12:50.042474 > COMPLETE!
21:12:50.042483 > SIO CMD processed in 1 ms
21:12:50.101183 > 
21:12:50.101198 > CF: 70 d8 03 02 4e
21:12:50.101707 > sioFuji::fujicmd_process() called, baud: 19200
EOF
output closed
join redirect
redirect joined
FujiNet ended with exit code 3221225477
```
*(Note: Exit code `3221225477` is hexadecimal `0xC0000005` = `STATUS_ACCESS_VIOLATION`)*.

---

### Root Cause Analysis

Tracing the issue through the firmware source code reveals why command `$D8` (`CMD::FUJI_COPY_FILE`) crashes on Atari SIO:

1. **Command Dispatch Conflict in `sioFuji::sio_process`:**
   In `lib/device/sio/sioFuji.cpp`:
   ```cpp
   void sioFuji::sio_process(const FujiSIOPacket &packet)
   {
       // Let the base class handle standard commands
       if (fujiDevice::processCommand(packet))
           return;

       switch (packet.command())
       {
       ...
       case CMD::FUJI_COPY_FILE:
           sio_copy_file(packet);
           break;
   ```
   `CMD::FUJI_COPY_FILE` is also registered in `fujiDevice::processCommand()` (`lib/device/fujiDevice/fujiDevice.cpp:140`):
   ```cpp
   { CMD::FUJI_COPY_FILE, [this](const FUJI_COMMAND_PACKET &packet) {
       uint8_t source = packet.param(0);
       uint8_t dest = packet.param(1);
       fujicmd_copy_file_success(source, dest, packet.dataAsString().value_or(""));
   } },
   ```
   Because `fujiDevice::processCommand(packet)` is evaluated first, it matches `CMD::FUJI_COPY_FILE` and returns `true`. Consequently, `sio_copy_file(packet)` in `sioFuji.cpp` is never reached.

2. **Atari SIO is a Multi-Phase Protocol:**
   On Atari SIO, `CMD::FUJI_COPY_FILE` is a two-phase transaction:
   - The initial 5-byte command frame (`CF: 70 d8 ...`) only conveys `sourceSlot` and `destSlot` in `aux1`/`aux2`.
   - The 256-byte copyspec payload (`source_path|dest_path`) is transmitted in a subsequent SIO data frame **after** FujiNet returns an ACK (`TRANS_STATE::WILL_GET`).
   
   Unlike buses where the entire packet payload is deserialized upfront before dispatch, `FujiSIOPacket` on SIO has not received the data payload at the time `sio_process` is called.

3. **Crash on Uninitialized `optional` in `FujiSIOPacket.h`:**
   When `fujiDevice` executes `packet.dataAsString()`:
   In `lib/bus/sio/FujiSIOPacket.h`:
   ```cpp
   const std::optional<ByteBuffer>& data() const {
       assert(_data.has_value());
       return _data;
   }
   const std::optional<const std::string> dataAsString() const {
       auto d = data();
       return std::string(reinterpret_cast<const char *>(d->data()), d->size());
   }
   ```
   Because no data frame has been fetched yet, `_data` is `std::nullopt`.
   - In debug builds, `assert(_data.has_value())` aborts execution.
   - In release builds (`NDEBUG`), `d->data()` calls `operator->()` on an empty `std::optional`, resulting in an immediate null pointer dereference (`0xC0000005` in `ucrtbase.dll`).

4. **Dead Code in `sioFuji::sio_copy_file`:**
   `sioFuji.cpp` already has a dedicated `sio_copy_file(packet)` function designed to handle the SIO data phase (`transaction_accept(TRANS_STATE::WILL_GET)` and `transaction_get(csBuf, sizeof(csBuf))`), but it is currently blocked by `fujiDevice::processCommand`.

---

### Suggested Solution

1. In `sioFuji::sio_process`, handle `CMD::FUJI_COPY_FILE` before delegating to `fujiDevice::processCommand(packet)` so that the SIO data phase can be accepted and read from the bus.
2. Once the 256-byte copyspec is received, delegate the copy to `fujicore_copy_file_success(source, dest, copySpec)`.
3. In `lib/bus/sio/FujiSIOPacket.h`, update `dataAsString()` to safely check `if (!_data.has_value()) return std::nullopt;` (matching `FujiBusPacket.h` in the rs232 implementation) to avoid crashes if called on unpopulated SIO packets.

RichStephens:
Fixes #1655

Tested with latest config in Fujisan.


FozzTexx:
Wait, why does this require changing CONFIG? I thought it was a firmware problem that #1687 fixes.

Also I didn't see a matching CONFIG commit for this. And when is someone going to make CONFIG automatically open a PR for disk changes?

RichStephens:
> Wait, why does this require changing CONFIG? I thought it was a firmware problem that #1687 fixes.
> 
> Also I didn't see a matching CONFIG commit for this. And when is someone going to make CONFIG automatically open a PR for disk changes?

I know, it's confusing.  There were two issues. One, a firmware issue where the copy failed in firmware.

Two, the config program itself was out of date and causing the host numbers being sent to the copy command to be wrong.  So, even when the test passed, the config copy didn't work.  The problem had already been fixed some time ago in the config program, but it had never been updated in the firmware.

FozzTexx (approved):

Issue #1689, opened 2026-09-23
[NDevice] NParser::read copies past receiveBuffer when READ asks for more than it holds


delysio:
`NParser::read()` resizes the output to `length` and copies `length` bytes from `_protocol->receiveBuffer` without checking its size. When the protocol's `read(length)` leaves fewer bytes there (for example `NetworkProtocolFS::read_file()` returns without touching the file when `receiveBuffer` is already non-empty), the copy runs past the end of the buffer and the computer gets bytes that were never received.

It is reachable over FTP today, where STATUS reports the socket count while READ is served from `receiveBuffer`; a PR fixing that count follows. Found by reading the code while chasing that bug, not reproduced on its own.

A fix would clamp the copy to `receiveBuffer->size()` and report the short read.

PR #1690, opened 2026-09-23
[network-protocol] Count receiveBuffer in FTP file available()

delysio:
# Summary

For an FTP file, `available()` counts only the socket and misses the bytes `status()` has already moved into `receiveBuffer`, so a client that reads what STATUS reports loses the end of the file. This makes the FILE branch report `receiveBuffer` first, as the DIR branch already does.

## Changes

- `NetworkProtocolFTP::available()`, FILE stream: report `receiveBuffer` while it holds data, otherwise the socket. Not the sum, because `read_file()` serves only one of them per READ. Shared `lib/network-protocol` code, so every platform reading FTP through N: is affected.

## Notes

- The socket-only count came in with #1138.
- Found in passing, not changed here: #1689.
- Clients that work around this by reading one byte at a time keep working.

## Testing

- Host build + ctest: `./build.sh -p ATARI` (10/10), `./build.sh -p COCO` (8/8).
- Firmware build: `fujinet-atari-v1` builds (PlatformIO 6.2, espressif32 6.10.0, ESP-IDF 5.4; `pio run -e fujinet-atari-v1` on the INI `build.sh -s` generates). All boards (`./build.sh -a`): not run; only the autobuild targets cover the rest.
- Flashed on hardware: no.
- Tests in `tests/`: none added; no ctest covers FTP.
- FujiNet-PC ATARI with Altirra over NetSIO, SpartaDOS X, a plain "STATUS, then READ what it reports" loop, files from ftp.amigascne.org compared byte for byte:

| file | master 1ee649d | this PR |
|---|---|---|
| mirror.txt, 1722 B | 0 bytes | identical |
| atop-0.7.1.lha, 30658 B | 26278 B, differs from byte 14600 | identical |
| P60A.LHA, 140141 B | not run | identical, 188 s |

A byte-at-a-time client needs 2366 s for P60A.LHA on master and still gets an identical file with this PR. Also checked against a local FTP server throttled to 20 KB/s and 8 KB/s, with a delayed start, an empty file and a missing file.

**Expected:** the plain loop receives the whole file; STATUS reports 0 waiting with the connection closed only after the last byte.

**Not tested:** not flashed on hardware, ESP32 built for `fujinet-atari-v1` only, no platform other than Atari exercised, FTP upload not exercised (unchanged).


RichStephens:
@tschak909 @FozzTexx Does this PR make sense to you?

I've devised an f-l-e test for this (https://github.com/RichStephens/fujinet-lib-experimental/tree/pr1690-ftp-test)
, and can confirm it fails, but my firmware repo is in a bit of a state right now and it's hard for me to pull this branch to prove it fixes it.  What I'm not clear on is if the fix is really needed, what human-observable behavior it fixes, and what existing behavior it might break

FozzTexx:
>  but my firmware repo is in a bit of a state right now and it's hard for me to pull this branch to prove it fixes it. 

You could always make a fresh clone in a new directory just for this PR then delete it when you're done. (And yes I too constantly forget I can do this. Great minds think alike!)

> What I'm not clear on is if the fix is really needed, what human-observable behavior it fixes, and what existing behavior it might break

It's the same fix that's had to be applied to pretty much every other class in `lib/network-protocol`, as a result of the current read and buffering design.

> I've devised an f-l-e test for this (https://github.com/RichStephens/fujinet-lib-experimental/tree/pr1690-ftp-test) , and can confirm it fails,

I was going to suggest a test, but because this problem potentially exists on all protocols, that would mean each protocol needs a similar test. And then what happens to the test when more protocols are added?

delysio:
> What I'm not clear on is if the fix is really needed, what human-observable behavior it fixes, and what existing behavior it might break.

Thanks. In short:

Behaviour. `status()` moves socket data into `receiveBuffer` before `available()` is read. Since #1138 the FTP FILE branch counts only the socket, so at the end of a transfer STATUS reports 0 bytes waiting and disconnected while the tail of the file is still in `receiveBuffer`. A client that reads what STATUS reports stops there, with no error:

| file | master | this PR |
|---|---|---|
| TEMP.TXT, 545 B | 0 B | identical |
| TEMP2.TXT, 11,600 B | 0 B | identical |
| atop-0.7.1.lha, 30,658 B | 26,278 B, corrupt from byte 14,600 | identical |

Today only reading one byte at a time works, about 12x slower.

**Risk.** Before #1138 FTP inherited `NetworkProtocolFS::available()`, which counted `receiveBuffer`; HTTP and the FTP DIR branch count it too. The count is the buffer or the socket, never the sum, so it never exceeds what the next READ returns (no path into #1689). Byte-at-a-time clients still get identical files.

RichStephens (approved):

